PR #68, opened 2018-05-12
Lint integration

kammce:
Because there are no readily available simple to use customizable C++ static code linter, I have decided on using very strict warnings using GCC warning arguments and using code review as a means of making sure code fits standard.


kammce:
cpplint by Google is not customizable therefore it is not a good candidate for this project.

PR #107, opened 2018-07-17
UART API

Jeremy-Chau:
uart.hpp with UART interface class created.


Jeremy-Chau:
There's 3 commits here because I accidentally created a second file so I deleted it.

PR #109, opened 2018-07-17
API for PWM

delwin-lei:
PWMInterface class with all functions that will be implemented.

### DIFF
--- a/firmware/library/L1_Drivers/pwm.hpp
+++ b/firmware/library/L1_Drivers/pwm.hpp
@@ -1,0 +1,13 @@
+#pragma once
+
+#include <cstdint>
+
+class PwmInterface
+{
+ public:
+    virtual void Initialize(uint32_t frequencyHz = 1'000) = 0;
+    virtual void SetDutyCycle(float dutyCycle) = 0;
+    virtual float GetDutyCycle() = 0;
+    virtual void SetFrequency(uint32_t frequencyHz) = 0;
+    virtual uint32_t GetFrequency() = 0;
+};


firmware/library/L1_Drivers/pwm.hpp
```diff
@@ -0,0 +1,18 @@
+#pragma once
+
+#include <cstdint>
+#include <cstdio>
+#include "L0_LowLevel/LPC40xx.h"
+#include "L1_Drivers/pin_configure.hpp"
```
kammce: cstdio, pin_configure, and LPC40xxx are not used here, they shouldn't be included.

firmware/library/L1_Drivers/pwm.hpp
```diff
@@ -0,0 +1,18 @@
...
+class PWMInterface
+{
+ public:
+        virtual void Initialize(uint32_t frequencyHz = 100) = 0;
+        virtual void SetDutyCycle(float dutyCycle) = 0;
+        virtual float GetDutyCycle() = 0;
+        virtual void SetFrequency(uint32_t frequencyHz) = 0;
+        virtual uint32_t GetFrequency() = 0;
+        virtual void SetAsDoubleEdge(bool vIsDoubleEdge = true) = 0;
+        virtual bool IsDoubleEdge() = 0;
```
kammce: Honestly, we can omit the implementation of double edge for now.

firmware/library/L1_Drivers/pwm.hpp
```diff
@@ -13,6 +10,4 @@ class PWMInterface
         virtual float GetDutyCycle() = 0;
```
kammce: Aren't you missing a `SetDutyCycle()` function?

kammce: Also, it would be great if you have one that accepts a float.

firmware/library/L1_Drivers/pwm.hpp
```diff
@@ -1,9 +1,6 @@
 #pragma once
 
 #include <cstdint>
-#include <cstdio>
-#include "L0_LowLevel/LPC40xx.h"
-#include "L1_Drivers/pin_configure.hpp"
 
 class PWMInterface
 {
```
kammce: I would think that you would also need an `Initialize()` function as well. Bare in mind that constructors should not be altering hardware.

firmware/library/L1_Drivers/pwm.hpp
delwin-lei: I believe I have that function on line 9.

firmware/library/L1_Drivers/pwm.hpp
delwin-lei: I have this function too on line 8.

firmware/library/L1_Drivers/pwm.hpp
```diff
@@ -0,0 +1,13 @@
+#pragma once
+
+#include <cstdint>
+
+class PWMInterface
```
kammce: Class name does not conform to Google C++ style.

firmware/library/L1_Drivers/pwm.hpp
```diff
@@ -0,0 +1,13 @@
...
+{
+ public:
+    virtual void Initialize(uint32_t frequencyHz = 100) = 0;
```
kammce: Suggestion: Make this 1000kHz. Seems like a good default for LED dimming.

PR #110, opened 2018-07-17
added piezo buzzer interface to L3_HAL directory

Hashah1:


firmware/library/L3_HAL/Piezo_Driver.cpp
```diff
@@ -0,0 +1,2 @@
+// cpp file for Piezo_Driver.hpp
```
kammce: You don't need a .cpp file if it isn't doing anything.

firmware/library/L3_HAL/Piezo_Driver.hpp
```diff
@@ -0,0 +1,39 @@
+// This driver is a layout for the piezo buzzer controller implementation
+// It contains the interface, which is inherited by the "Buzzer" class
+// which implements the functions in the interface
+// The constructor is responsible for initializing the port
+
+#pragma once
+
+// import only the necessary libraries to make this interface function
```
kammce: This is something you do, but you don't need to add this comment.

firmware/library/L3_HAL/Piezo_Driver.hpp
```diff
@@ -0,0 +1,39 @@
...
+#include <cstdint>
+
+// import the lpc_pwm class from SJOne library as our PWM driver
```
kammce: Since it is clear that you are importing this file, no need to comment that you are including this file

firmware/library/L3_HAL/Piezo_Driver.hpp
```diff
@@ -0,0 +1,39 @@
...
+#include "/Users/mianhashimshah/OpenSJ_Bluz/SJOne/L2_Drivers/lpc_pwm.hpp"
```
kammce: This will break if you include the file like this. Don't use absolute directory paths that include specifics about your directory.

firmware/library/L3_HAL/Piezo_Driver.hpp
```diff
@@ -0,0 +1,39 @@
...
+
+// BuzzerInterface is inherited by Buzzer class
```
kammce: Remove all obvious comments.

firmware/library/L3_HAL/Piezo_Driver.hpp
```diff
@@ -0,0 +1,39 @@
...
+class BuzzerInterface
+{
+ public:
+        // The init() class is responsible for selecting the pin configuration
+        virtual void init(uint8_t pinNumber) = 0;
```
kammce: Spell this out fully and follow the google naming standard. 
See: https://google.github.io/styleguide/cppguide.html#Naming

firmware/library/L3_HAL/Piezo_Driver.hpp
```diff
@@ -0,0 +1,39 @@
...
+
+        // This is responsible to change the volume
+        // of the buzzer via the duty cycle
+        // 50% is max volume; < 50% is a lower volume
+        virtual void setVolume(uint32_t dutyCycle) = 0;
```
kammce: The volume should be a float and between 0 and 1. The user shouldn't have to know that setting volume requires them to set the duty cycle directly.

firmware/library/L3_HAL/Piezo_Driver.hpp
```diff
@@ -0,0 +1,39 @@
...
+
+        // Stops the buzzer from making any sound
+        virtual void stop() = 0;
+
+        // beep continuously for a certain period of time
+        // @ a specified frequency
+        virtual void beep(/*uint32_t frequency*/, uint32_t duration) = 0;
```
kammce: You should include the frequency in these methods.

firmware/library/L3_HAL/Piezo_Driver.hpp
```diff
@@ -0,0 +1,31 @@
...
+        // beep continuously for a certain period of time
+        // @ a specified frequency
+        virtual void Beep(uint32_t frequency, uint32_t duration) = 0;
+
+        // produce a sequence of beeps for a specified duration
+        // @ a specified frequency
+        virtual void ProduceSequence(uint32_t freq, uint32_t dur) = 0;
+
+        // produce a musical tone for a specified duration
+        virtual void MakeTone(uint32_t duration) = 0;
```
kammce: I think that having the two functions, `Beep` and `MakeTone` is unnecessary. You can combine them into 1 
function. They do the same thing thus you can write the method declaration as:

    virtual void Beep(uint32_t duration, uint32_t frequency = 1'000)

This would eliminate the need for MakeTone. Or you could name it MakeTone and remove Beep. Which ever name you feel fits the best.

I'm still now sure what ProduceSequence does. Can you explain how it works?

PR #114, opened 2018-07-18
GPIO API Interface

PricelessAntonio:
Initial commit to the unstable GPIO-API branch

### DIFF
--- a/firmware/library/L1_Drivers/gpio.hpp
+++ b/firmware/library/L1_Drivers/gpio.hpp
@@ -1,0 +1,23 @@
+#pragma once
+
+class GpioInterface
+{
+ public:
+    enum PinDirection : uint8_t
+    {
+        kInput  = 0,
+        kOutput = 1
+    };
+    enum PinOutput : uint8_t
+    {
+        kLow  = 0,
+        kHigh = 1
+    };
+    virtual void SetAsInput(void)                                    = 0;
+    virtual void SetAsOutput(void)                                   = 0;
+    virtual void SetDirection(GpioInterface::PinDirection direction) = 0;
+    virtual void SetHigh(void)                                       = 0;
+    virtual void SetLow(void)                                        = 0;
+    virtual void Set(GpioInterface::PinOutput output = kHigh)        = 0;
+    virtual bool ReadPin(void)                                       = 0;
+};


firmware/library/L1_Drivers/gpio.hpp
```diff
@@ -0,0 +1,67 @@
...
+     virtual void SetAsInput(void)             = 0;
+     virtual void SetAsOutput(void)            = 0;
+     virtual void SetDir(bool input = true)    = 0;
+     virtual void CtrlOutputHigh(void)         = 0;
+     virtual void CtrlOutputLow(void)          = 0;
+     virtual void CtrlOutput(bool high = true) = 0;
+     virtual bool ReadPin(void)                = 0;
+}
+
+class GPIO : public GPIOInterface, public PinConfigure
```
kammce: Do not include implementation of GPIO in this commit.

firmware/library/L1_Drivers/gpio.hpp
```diff
@@ -0,0 +1,67 @@
...
+
+class GPIOInterface
+{
+ public:
+     virtual void SetAsInput(void)             = 0;
+     virtual void SetAsOutput(void)            = 0;
+     virtual void SetDir(bool input = true)    = 0;
+     virtual void CtrlOutputHigh(void)         = 0;
+     virtual void CtrlOutputLow(void)          = 0;
+     virtual void CtrlOutput(bool high = true) = 0;
```
kammce: I think that replacing `Ctrl` with `Set` would be more readable. Ctrl seems kinda confusing. Set was also make more sense in place of `CtrlOutput` especially since it sets the pin to high. So, `SetOutput()` can either mean, set it to the input parameter or set it (the opposite of clear) the pin.

PricelessAntonio: Is it alright if I just change the names to SetHigh(), SetLow(), and Set(bool high = true)?  I like those names better because users would not have to write a long method name when using those methods.

kammce: Thats good!

firmware/library/L1_Drivers/gpio.hpp
```diff
@@ -0,0 +1,17 @@
...
+#include <stdint.h>
+
+#include "L0_LowLevel/LPC40xx.h"
+
+class GpioInterface
+{
+ public:
+     virtual void SetAsInput(void)          = 0;
+     virtual void SetAsOutput(void)         = 0;
+     virtual void SetDir(bool input = true) = 0;
```
kammce: Prefer that method and variable names be spelled out in full. This should be `SetDirection`.

kammce: Also, looking at the input parameter, having a bool doesn't exactly read well. For example, `gpio.SetDirection(true);` doesn't immediately tell the user what is happening. True could mean output... or input, its a bit ambiguous. And setting it to default to true, means developers can do `gpio.SetDirection()`, but the behavior is still ambiguous.

It would be useful to create an `enum class` with constants such as, `kInput` and `kOutput`. See PinConfigInterface for an example: https://github.com/kammce/SJSU-Dev2/blob/master/firmware/library/L1_Drivers/pin_configure.hpp

firmware/library/L1_Drivers/gpio.hpp
```diff
@@ -0,0 +1,17 @@
+#pragma once
+
+#include <stdint.h>
+
+#include "L0_LowLevel/LPC40xx.h"
```
kammce: This isn't used, so it should be removed.

firmware/library/L1_Drivers/gpio.hpp
```diff
@@ -0,0 +1,17 @@
+#pragma once
+
+#include <stdint.h>
```
kammce: This isn't used, so it should be removed.

kammce: Also, standard states that you must use the C++ include of stdint.h which is `<cstdint>`

firmware/library/L1_Drivers/gpio.hpp
PricelessAntonio: Would the Set method need a enum class as well, or is it intuitive enough to understand? (i.e. Set(true) means set high and Set(false) means set low)

firmware/library/L1_Drivers/gpio.hpp
PricelessAntonio: Just to be clear, #pragma once is left in for this commit, but #include <stdint.h> is left out since it is not being used at the moment, right?

firmware/library/L1_Drivers/gpio.hpp
PricelessAntonio: Also, should I just take out the SetDirection method?  If the user has to write a function that long that has restricted input in the form of enums, they could just write SetAsInput() or SetAsOutput(), which is shorter in written length.  The only benefit I see to having the SetDirection() method would be if it took in a boolean parameter, allowing users to put functions within the method in order to change the direction during runtime.  Even then, I don't think users would need to change the direction of the GPIO pin multiple times.  Any thoughts on that?

firmware/library/L1_Drivers/gpio.hpp
PricelessAntonio: Also when i start to use cstdint, do I use the scope resolution operator for uint8_t?

kammce: Nope, cstdint is the same as stdint.h. No need for the scope operator.

kammce: You need #pragma once, to be your include guard. You don't use anything from stdint here so remove that include.

firmware/library/L1_Drivers/gpio.hpp
kammce: I see your point. Alright, create a enum, and have the parameter be that enum type. Do not use a strongly typed "enum class", so that the user can use 0 (input) or 1 (output) as a means of changing directions.

firmware/library/L1_Drivers/gpio.hpp
```diff
@@ -0,0 +1,24 @@
...
+        kOutput = 1
+    };
+    enum PinOutput : uint8_t
+    {
+        kLow  = 0,
+        kHigh = 1
+    };
+    virtual void SetAsInput(void)                             = 0;
+    virtual void SetAsOutput(void)                            = 0;
+    virtual void SetDirection(
```
kammce: Don't have a default of this. It doesn't read right if the user doesn't supply a parameter. Keep the enum though.

PR #115, opened 2018-07-18
SSP API Interface

diarmadhi:
SSP API interface for SJSU-Dev2.

### DIFF
--- a/firmware/library/L1_Drivers/ssp.hpp
+++ b/firmware/library/L1_Drivers/ssp.hpp
@@ -1,0 +1,71 @@
+// SSP provides the ability for serial communication over SPI, SSI, or
+// Microwire on the LPC407x chipset. NOTE: The SSP2 peripheral is not
+// available on the SJTwo board, as such this driver does not support
+// it. Only one set of pins are available for either SSP0 or SSP1 as
+// follows:
+//      Peripheral  |   MISO    |   MOSI    |   SCK
+//          SSP0    |   P0.17   |   P0.18   |   P0.15
+//          SSP1    |   P0.8    |   P0.9    |   P0.7
+#pragma once
+
+class SspInterface
+{
+ public:
+     // SSP peripherals
+    enum class Peripheral
+    {
+        kSsp0 = 0,
+        kSsp1 = 1
+    };
+
+    enum PowerOn
+    {
+        kPconp0 = 21,
+        kPconp1 = 10
+    };
+
+    // SSP frame formats
+    enum FrameMode
+    {
+        kSpi     = 0,
+        kTi      = 1,
+        kMicro   = 2
+    };
+
+    // SSP Master/slave modes
+    enum MasterSlaveMode
+    {
+        kMaster  = 0,
+        kSlave   = 1
+    };
+
+
+    // SSP data size for frame packets
+    enum DataSize
+    {
+        kFour        = 0b0011,  // 4-bit  transfer
+        kFive        = 0b0100,  // 5-bit  transfer
+        kSix         = 0b0101,  // 6-bit  transfer
+        kSeven       = 0b0110,  // 7-bit  transfer
+        kEight       = 0b0111,  // 8-bit  transfer
+        kNine        = 0b1000,  // 9-bit  transfer
+        kTen         = 0b1001,  // 10-bit transfer
+        kEleven      = 0b1010,  // 11-bit transfer
+        kTwelve      = 0b1011,  // 12-bit transfer
+        kThirteen    = 0b1100,  // 13-bit transfer
+        kFourteen    = 0b1101,  // 14-bit transfer
+        kFifteen     = 0b1110,  // 15-bit transfer
+        kSixteen     = 0b1111,  // 16-bit transfer
+    };
+
+    virtual bool Initialize()                                       = 0;
+    virtual bool GetTransferStatus()                                = 0;
+    virtual uint8_t Transfer(uint16_t data)                         = 0;
+    virtual void SetPeripheralPower()                               = 0;
+    virtual void SetPeripheralMode(
+            MasterSlaveMode mode, FrameMode frame, DataSize size)   = 0;
+    // NOTE: clock_divider is the number of prescaler-output clocks per bit.
+    // See page 611 of UM10562 for SCR-Serial Clock Rate for more info.
+    virtual void SetClock(bool polarity, bool phase,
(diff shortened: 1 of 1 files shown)


firmware/library/L1_Drivers/ssp.hpp
```diff
@@ -0,0 +1,206 @@
+/**
+ ** Provide How-to for driver
```
kammce: Please place a TODO here.
The todo should look like this:
`// TODO(#1241) Provide How-to for driver`

Where #1241 is the issue number

Also, only use // for comments including multiline comments

firmware/library/L1_Drivers/ssp.hpp
```diff
@@ -0,0 +1,206 @@
+/**
+ ** Provide How-to for driver
+
+ **
+ **
+ */
+
```
kammce: nit: remove newline seperator

firmware/library/L1_Drivers/ssp.hpp
```diff
@@ -0,0 +1,206 @@
...
+ ** Provide How-to for driver
+
+ **
+ **
+ */
+
+#pragma once
+
+#include "LPC40xx.h"
+#include "pin_configure.hpp"
```
kammce: Fully qualify the path with respect to the library folder. For example, this should be "L1_Driver/pin_configure.hpp"

firmware/library/L1_Drivers/ssp.hpp
```diff
@@ -0,0 +1,206 @@
...
+    // SSP Master/slave modes
+    enum MSMode
+    {
+        MASTER  = 0;
+        SLAVE   = 1
+    };
+
+
+    // SSP data size for frame packets
+    enum DataSize
```
kammce: For these to actually be binary, you need to prefix them with `0b`. For example FOUR should be `0b0011`. Otherwise your number is eleven.

firmware/library/L1_Drivers/ssp.hpp
```diff
@@ -0,0 +1,206 @@
...
+    {
+        FOUR    = 0011,      // 4-bit  transfer
+        FIVE    = 0100,      // 5-bit  transfer
+        SIX     = 0101,      // 6-bit  transfer
+        SEVEN   = 0110,      // 7-bit  transfer
+        EIGHT   = 0111,      // 8-bit  transfer
+        NINE    = 1000,      // 9-bit  transfer
+        TEN     = 1001,      // 10-bit transfer
+        ELEVEN  = 1010,      // 11-bit transfer
+        TWELVE  = 1011,      // 12-bit transfer
```
kammce: Equals should be aligned with each other for this whole enum series.

firmware/library/L1_Drivers/ssp.hpp
```diff
@@ -0,0 +1,206 @@
...
+        FIFTEEN     = 1110,  // 15-bit transfer
+        SIXTEEN     = 1111,  // 16-bit transfer
+    };
+
+    virtual bool init()                                         = 0;
+    virtual bool get_transfer_status(Peripheral sspx)           = 0;
+    virtual uint8_t transfer(Peripheral sspx, uint8_t data)     = 0;
+};
+
+class SSP : SSPInterface
```
kammce: Only include the interface, do not include the implementation for this commit.

firmware/library/L1_Drivers/ssp.hpp
kammce: This should also be just 1 line for now.

kammce:
Please rename this pull request and change your commit to not include information like, "I made a branch to". The information about branches or even your repo is irrelevant for the pull request.

firmware/library/L1_Drivers/ssp.hpp
diarmadhi: Is the Issue number an arbitrary number, or is #1241 a specific issue # that I'm not seeing right now?

kammce: Its whatever git assigns the number to be.

firmware/library/L1_Drivers/ssp.hpp
```diff
@@ -0,0 +1,61 @@
+// TODO(#1241) Provide How-to for Driver
+
+#pragma once
+
+#include "L0_LowLevel/LPC40xx.h"
+#include "L1_Drivers/pin_configure.hpp"
```
kammce: Remove headers that are not used (both of them).

firmware/library/L1_Drivers/ssp.hpp
```diff
@@ -0,0 +1,61 @@
...
+    };
+
+    // SSP Master/slave modes
+    enum MSMode
+    {
+        MASTER  = 0;
+        SLAVE   = 1
+    };
+
+
```
kammce: Remove extra newline.

firmware/library/L1_Drivers/ssp.hpp
kammce: Since DataSize is not used here, it shouldn't be in the interface. But I am wondering why there isn't a method in the interface that uses this enum. If looks like you are missing quite a few methods for your interface.

diarmadhi: As far as the enums go, they are all being used by the struct in the implementation which in turn is used by init(). I think I may have to move all the enums to the implementation. I'm also planning on adding getter and setter functions, which I'm assuming should be added to the interface as well?

kammce: Yes, define them in the implementation class, but not here if they aren't used here. But, if you are adding setters and getters that use these, then that is fine to keep them there.

firmware/library/L1_Drivers/ssp.hpp
```diff
@@ -0,0 +1,25 @@
+// SSP provides the ability for serial communication over SPI, SSI, or
+// Microwire on the LPC407x chipset. NOTE: The SSP2 peripheral is not
+// available on the SJTwo board, as such this driver does not support
+// it. Only one set of pins are available for either SSP0 or SSP1 as
+// follows:
+//      Peripheral  |   MISO    |   MOSI    |   SCK
+//          SSP0    |   P0.17   |   P0.18   |   P0.15
+//          SSP1    |   P0.8    |   P0.9    |   P0.8
```
diarmadhi: Need to change port pin for SSP1 SCK to P0.7

PR #117, opened 2018-07-18
Initial commit for UART API

Jeremy-Chau:


firmware/library/L1_Drivers/uart.hpp
```diff
@@ -0,0 +1,11 @@
+#pragma once
+
+class UARTInterface
+{
+ public:
+    virtual void setBaud(unsigned int baudrate) = 0;
```
kammce: Follow Google C++ style guide. Should be `SetBaud`.
`int` has different size depending on what system you are compiling for, please use <cstdint>

firmware/library/L1_Drivers/uart.hpp
```diff
@@ -0,0 +1,11 @@
...
+    virtual bool init(unsigned int baud, uint32_t queue_size = 16) = 0;
+    virtual void IntHandle() = 0;
```
kammce: Write the full name of the function and what it is doing. A function or method should say and action and not be a nown. For example, would prefer `AttachInterruptHandler(...)`

Also, the method should accept a function pointer.

firmware/library/L1_Drivers/uart.hpp
```diff
@@ -0,0 +1,11 @@
+#pragma once
+
+class UARTInterface
+{
+ public:
+    virtual void setBaud(unsigned int baudrate) = 0;
+    virtual bool init(unsigned int baud, uint32_t queue_size = 16) = 0;
```
kammce: Prefer to spell words out fully, for example `Initialize`

firmware/library/L1_Drivers/uart.hpp
```diff
@@ -0,0 +1,11 @@
...
+    virtual void IntHandle() = 0;
+    virtual void send(char out, unsigned int time_limit) = 0;
```
kammce: Ditto from the above comment about stdint and naming.

firmware/library/L1_Drivers/uart.hpp
```diff
@@ -0,0 +1,11 @@
+#pragma once
+
+class UARTInterface
```
kammce: See the naming rules for classes here: https://google.github.io/styleguide/cppguide.html#General_Naming_Rules

kammce:
Please squash your commits into 1.

firmware/library/L1_Drivers/uart.hpp
```diff
@@ -0,0 +1,13 @@
+#pragma once
+
+#include <cstdint>
+
+class UART_Interface
```
kammce: No underscore.

firmware/library/L1_Drivers/uart.hpp
```diff
@@ -0,0 +1,13 @@
...
+{
+ public:
+    virtual void SetBaud(uint32_t baudrate) = 0;
+    virtual bool Initialize(uint32_t baud, uint32_t queue_size = 16) = 0;
+    virtual void InteruptHandler(void) = 0;
```
kammce: What does this do? I still can't figure out what  UART's would do with this implemented method.

firmware/library/L1_Drivers/uart.hpp
```diff
@@ -0,0 +1,13 @@
+#pragma once
+
+#include <cstdint>
+
+class UART_Interface
+{
+ public:
+    virtual void SetBaud(uint32_t baudrate) = 0;
+    virtual bool Initialize(uint32_t baud, uint32_t queue_size = 16) = 0;
```
kammce: You should also provide methods for setting queue_size.

firmware/library/L1_Drivers/uart.hpp
```diff
@@ -0,0 +1,13 @@
...
+    virtual void InteruptHandler(void) = 0;
+    virtual void SendData(char out, uint32_t time_limit) = 0;
+    virtual bool ReceiveData(char *CharInput, uint32_t time_limit) = 0;
```
kammce: variables (which includes parametes) should be snake_case.

kammce: Prefer just the word `Send` and just `Recieve`

firmware/library/L1_Drivers/uart.hpp
Jeremy-Chau: This method would be called whenever an interrupt is created due to the Rx pin receiving data. The method would take that data from the pin and throw it onto the queue waiting to be picked up by the user. The user would then use the receive function to grab that data whenever they need to.

firmware/library/L1_Drivers/uart.hpp
Jeremy-Chau: You stated during the design doc phase that you didn't want the queue_size to be dynamically set. This is why I didn't provide a method to set the queue size before. Did you want to have queue_size dynamically set now?

kammce: My original thought was that you were creating a queue, but now I see this refers to the FIFO (correct me if I'm wrong). Also, looking at the datasheet, I don't see a register that allows you to change the FIFO length. If that is the case, then remove queue_size as an option.

firmware/library/L1_Drivers/uart.hpp
kammce: I dont thing that the interface should include an InterruptHandler. You cannot attach an interrupt to a class's method unless it is static. Therefor, there would need to be a templated static handler for the various UART types. So this method should be removed.

firmware/library/L1_Drivers/uart.hpp
Jeremy-Chau: What if the queue_size parameter is used for a FreeRTOS queue being initialized in the function?

PR #122, opened 2018-07-21
Utility Namespace API

hangbogu:


firmware/library/L2_Utilities/util.hpp
```diff
@@ -0,0 +1,28 @@
...
+
+class util
+{
+ private:
+    util();
+
+ public:
+    static void delayUS(uint64_t delay_time);
+    static void delayMS(uint64_t delay_time);
+    static void logBootInfo(char destination_buffer[256]);
```
kammce: You can omit this. No need for this function to be in here.

firmware/library/L2_Utilities/util.hpp
```diff
@@ -0,0 +1,28 @@
...
+    static uint64_t micros();
+    static uint32_t millis();
+    static uint32_t random(uint32_t min, uint32_t max);
+    static void randomSeed();
```
kammce: The should be a verb phrase, like SetRandomSeed.

firmware/library/L2_Utilities/util.hpp
```diff
@@ -0,0 +1,28 @@
...
+    static uint32_t bitExtract(uint32_t var, uint8_t start_bit, uint8_t width);
+    static void bitPlace(uint32_t * var, uint8_t start_bit, uint8_t width);
```
kammce: On second thought, it would better for these to take a 32-bit number and return the changed value, rather than changing it in place.

kammce: Same goes for the functions below.

firmware/library/L2_Utilities/util.hpp
```diff
@@ -0,0 +1,28 @@
...
+    static void bitSet(uint32_t * var, uint8_t bit);
+    static void bitClear(uint32_t * var, uint8_t bit);
+    static uint32_t bitRead(uint32_t var, uint8_t bit);
+    static uint32_t Min(uint32_t var_1, uint32_t var_2);
+    static uint32_t Max(uint32_t var_1, uint32_t var_2);
+    static uint32_t map(uint32_t value, uint32_t cur_low, uint32_t cur_high,
+                        uint32_t new_low, uint32_t new_high);
+    static uint32_t constrain(uint32_t in, uint32_t low, uint32_t high);
+};
```
kammce: All methods should be capitalized as per the Google C++ style guide.

firmware/library/L2_Utilities/util.hpp
kammce: Also see the version of this in L1_driver/pin_configure.hpp method BitPlace. Make sure to name your variables something that can be easily understood. For example, I"m not sure what var means in this case. Everything is a variable.

firmware/library/L2_Utilities/util.hpp
```diff
@@ -0,0 +1,23 @@
+#pragma once
+
+#include <cstdint>
+
+namespace Utility
+{
+static uint32_t Random(uint32_t min, uint32_t max);
+static void SetRandomSeed();
```
kammce: I think that `SetRandomSeed()` should allow the developer specify a random seed input. 
You can have it default to 0xFFFFFFFF, in that case, you can use the <ctime> library in the implementation.

firmware/library/L2_Utilities/util.hpp
```diff
@@ -0,0 +1,23 @@
+#pragma once
+
+#include <cstdint>
+
+namespace Utility
+{
+static uint32_t Random(uint32_t min, uint32_t max);
```
kammce: Remove static from all of the method definitions. Static used in this way limits the function calls to the scope of this file.

firmware/library/L2_Utilities/util.hpp
```diff
@@ -0,0 +1,23 @@
...
+static void SetRandomSeed();
+static uint32_t BitExtract(uint32_t extractee, uint8_t start_bit,
```
kammce: Would prefer the word target, `extractee` doesn't seem as readable. That goes for the other implementations as well.

firmware/library/L2_Utilities/util.hpp
```diff
@@ -0,0 +1,23 @@
...
+                           uint8_t width);
+static uint32_t BitPlace(uint32_t placee, uint8_t s_bit, uint8_t bits,
```
kammce: Do not short hand variable names. Spell out the whole word/phrase so it is readable.

firmware/library/L2_Utilities/util.hpp
```diff
@@ -0,0 +1,23 @@
...
+                         uint8_t width);
+static uint32_t BitPlace(uint32_t placee, uint8_t s_bit, bool bit,
+                         uint8_t width);
+static uint32_t BitSet(uint32_t setee, uint8_t bit);
+static uint32_t BitClear(uint32_t clearee, uint8_t bit);
+static uint32_t BitRead(uint32_t readee, uint8_t bit);
```
kammce: Would prefer bit_position, since the meaning of this `bit` variable and the above `bool bit` are different. This is to reduce confusion.

firmware/library/L2_Utilities/util.hpp
```diff
@@ -0,0 +1,23 @@
...
+static uint32_t Min(uint32_t comp_1, uint32_t comp_2);
```
kammce: Please write variable names in full.

firmware/library/L2_Utilities/util.hpp
```diff
@@ -0,0 +1,23 @@
...
+static uint32_t Max(uint32_t comp_1, uint32_t comp_2);
+static uint32_t Map(uint32_t value, uint32_t cur_low, uint32_t cur_high,
+                    uint32_t new_low, uint32_t new_high);
+static uint32_t Constrain(uint32_t in, uint32_t low, uint32_t high);
```
kammce: would prefer the words min and max rather than low and high.

PR #127, opened 2018-07-24
Revison of Button API Interface

Lawrence-Wan:
Revised Interface of the button driver.
- Combined invert and default button signal methods into one method
   (InvertButtonSignal).
- Squashed to 1 commit.

### DIFF
--- a/firmware/library/L1_Drivers/button_driver.hpp
+++ b/firmware/library/L1_Drivers/button_driver.hpp
@@ -1,0 +1,10 @@
+#pragma once
+
+class ButtonInterface {
+ public:
+         virtual void Init()                = 0;
+         virtual bool Read()                = 0;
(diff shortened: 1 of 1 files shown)


firmware/library/L1_Drivers/button_driver.hpp
```diff
@@ -0,0 +1,10 @@
+#pragma once
+
+class ButtonInterface {
+ public:
+         virtual void Init()                = 0;
+         virtual bool GetButtonState()      = 0;
```
kammce: I think `GetState` or simply `read()` might be better alternatives to `GetButtonState()`. Seems a bit verbose especially since the user will probably be doing something like this:
```
Button btn1;
btn1.GetButtonState();
```
Seems a bit repetitive.

PR #128, opened 2018-07-24
Integrating FreeRTOS

kammce:


firmware/examples/FreeRTOS_Blinker/source/main.cpp
```diff
@@ -0,0 +1,83 @@
...
+        vTaskDelay(50);
+    }
+}
+
+}  // namespace
+
+int main(void)
+{
+    DEBUG_PRINT("Creating Tasks ...");
+    xTaskCreate(vLedToggle, "vLedToggle", 256, NULL, tskIDLE_PRIORITY + 1,
```
MasterBoyd: Create const expressions for different priorities instrad of using tskIDLE_PRIORITY+1/+3/+etc..

firmware/examples/FreeRTOS_Blinker/source/main.cpp
```diff
@@ -0,0 +1,83 @@
...
+                NULL);
+    xTaskCreate(vButtonReader, "vButtonReader", 256, NULL, tskIDLE_PRIORITY + 2,
```
MasterBoyd: Comment out what each part of the task create stands for.

MasterBoyd: Comment out what each part of the task create stands for.

firmware/examples/FreeRTOS_Blinker/source/main.cpp
MasterBoyd: Create const expressions for different priorities instrad of using tskIDLE_PRIORITY+1/+3/+etc.

firmware/examples/FreeRTOS_Blinker/source/main.cpp
MasterBoyd: Use function that allocates the minimum amount of necessary stack space plus the stack space you need for your implementation.

PR #137, opened 2018-07-31
Pwm driver implementation

delwin-lei:


firmware/library/L1_Drivers/pwm.hpp
```diff
@@ -11,3 +14,129 @@ class PwmInterface
     virtual void SetFrequency(uint32_t frequency_hz) = 0;
     virtual uint32_t GetFrequency() = 0;
 };
+
+class Pwm : public PwmInterface
+{
+ public:
+    static constexpr uint8_t kPwmFunction = 0b001;
+
```
kammce: Remove extra line between constants.

firmware/library/L1_Drivers/pwm.hpp
```diff
@@ -11,3 +14,129 @@ class PwmInterface
...
+    static constexpr uint8_t kDefaultFunction = 0b000;
+
+    // remove when sysclock is complete
```
kammce: Should be a TODO. Make an issue for this as well.

firmware/library/L1_Drivers/pwm.hpp
```diff
@@ -11,3 +14,129 @@ class PwmInterface
...
+        static_assert(1 <= pwm_port && pwm_port <= 2,
+                      "Available PWM ports are only on port 1 or 2.");
+        static_assert(pwm_pin <= 5,
+                      "Available PWM pins only go up to pin 5.");
+        return Pwm(pwm_port, pwm_pin);
+    }
+
+    Pwm(uint8_t vpport, uint8_t vppin):
+        pwm_(pwm_pin_),
+        pwm_pin_(vpport, vppin)
```
kammce: Constructor should not modify hardware.

firmware/library/L1_Drivers/pwm.hpp
```diff
@@ -11,3 +14,129 @@ class PwmInterface
...
+    static Pwm CreatePwm()
+    {
+        static_assert(1 <= pwm_port && pwm_port <= 2,
+                      "Available PWM ports are only on port 1 or 2.");
+        static_assert(pwm_pin <= 5,
+                      "Available PWM pins only go up to pin 5.");
+        return Pwm(pwm_port, pwm_pin);
+    }
+
+    Pwm(uint8_t vpport, uint8_t vppin):
```
kammce: Make this a constexpr constructor.

firmware/library/L1_Drivers/pwm.hpp
```diff
@@ -11,3 +14,129 @@ class PwmInterface
...
+            port = 1;
+            channel = static_cast<uint8_t>(pwm_pin_.GetPin() + 1);
+        }
+        duty_cycle = 0;
+    }
+
+    ~Pwm()
+    {
+        // 0b000 is GPIO
+        pwm_pin_.SetPinFunction(kDefaultFunction);
```
kammce: This is an interesting construct here. I'm not sure if we should not allow it, or make it a standard that everyone follows. 
@SameerAzer @MasterBoyd take a look at this and tell me what you think.

firmware/library/L1_Drivers/pwm.hpp
```diff
@@ -11,3 +14,129 @@ class PwmInterface
...
+            pwm_pin_.SetPinFunction(kPwmFunction);
+            port = 1;
+            channel = static_cast<uint8_t>(pwm_pin_.GetPin() + 1);
+        }
+        duty_cycle = 0;
+    }
+
+    ~Pwm()
+    {
+        // 0b000 is GPIO
```
kammce: Rather than putting a comment here to explain what kDefaultFunction is, it would be better to simply rename kDefaultFunction to like kGpioFunction.

firmware/library/L1_Drivers/pwm.hpp
```diff
@@ -11,3 +14,129 @@ class PwmInterface
...
+        pwm_pin_.SetPinFunction(kDefaultFunction);
+    }
+
+    void Initialize(uint32_t frequency_hz = 1'000) override
+    {
+        // DIRECT MANIUPULATION OF SYSTEM CLOCK REGISTERS
```
kammce: Make this a TODO.

firmware/library/L1_Drivers/pwm.hpp
```diff
@@ -11,3 +14,129 @@ class PwmInterface
...
+        // WILL REPLACE WITH FUNCTIONS LATER
+        // Enables PWM1 power/clock control bit
+        LPC_SC->PCONP |= (1 << 6);
```
kammce: Define 6 as a constant. Prefer not to use magic numbers.

firmware/library/L1_Drivers/pwm.hpp
```diff
@@ -11,3 +14,129 @@ class PwmInterface
...
+        // input clock is divided by 1
+        LPC_SC->PCLKSEL = (1 << 0);
```
kammce: Prefer not to use magic numbers.

firmware/library/L1_Drivers/pwm.hpp
```diff
@@ -11,3 +14,129 @@ class PwmInterface
...
+        // END OF DIRECT MANIPULATION
+
+        match_value = kDefaultClockSpeed / frequency_hz;
+
+        // Resets PWMTC on Match with MR0
+        LPC_PWM1->MCR |= (1 << 1);
```
kammce: Prefer not to use magic numbers. Same goes for the rest...

firmware/library/L1_Drivers/pwm.hpp
```diff
@@ -11,3 +14,129 @@ class PwmInterface
...
+        LPC_PWM1->MR0 = match_value;
+
+        // Enables PWM TC and PC for counting and enables PWM mode
+        LPC_PWM1->TCR = (1<< 0) | (1 << 3);
+        LPC_PWM1->CTCR &= ~(0xF << 0);
+
+
+        // Enables PWM[channel] output
+        LPC_PWM1->LER |= ~(1 << 0);
+        LPC_PWM1->PCR |= (1 << (channel + 8));
```
kammce: Prefer the `channel + 8` to be a private constexpr method.

firmware/library/L1_Drivers/pwm.hpp
```diff
@@ -11,3 +14,129 @@ class PwmInterface
...
+    }
+
+    void SetDutyCycle(float percentage) override
+    {
+        if (percentage < 0 || percentage > 100)
+        {
```
kammce: Add TODO comment to replace this with an `assert`.

firmware/library/L1_Drivers/pwm.hpp
```diff
@@ -11,3 +14,129 @@ class PwmInterface
...
+            return;
+        }
+        duty_cycle = static_cast<uint32_t>((percentage * static_cast<float>(match_value)) / 100);
+
+        switch (channel)
```
kammce: Would prefer making this into a look up table.

firmware/library/L1_Drivers/pwm.hpp
```diff
@@ -11,3 +14,129 @@ class PwmInterface
...
+ private:
+    PinConfigureInterface & pwm_;
+
+    PinConfigure pwm_pin_;
+
+    static uint32_t match_value;
+
+    // will signify current PWM0/1 and channel
+    bool port;
+    uint8_t channel;
```
kammce: make this const since it doesn't change.

firmware/library/L1_Drivers/pwm.hpp
```diff
@@ -11,3 +14,129 @@ class PwmInterface
...
+
+ private:
+    PinConfigureInterface & pwm_;
+
+    PinConfigure pwm_pin_;
+
+    static uint32_t match_value;
+
+    // will signify current PWM0/1 and channel
+    bool port;
```
kammce: make this const.

firmware/library/L1_Drivers/pwm.hpp
```diff
@@ -11,3 +14,129 @@ class PwmInterface
...
+    {
+        return kDefaultClockSpeed / match_value;
+    }
+
+ private:
+    PinConfigureInterface & pwm_;
+
+    PinConfigure pwm_pin_;
+
+    static uint32_t match_value;
```
kammce: You should make an getter for this match value just in case you need it for testing.

firmware/library/L1_Drivers/pwm.hpp
kammce: Also, add a constructor that accepts a PinConfigure & object for dependency injection.

firmware/library/L1_Drivers/pwm.hpp
```diff
@@ -11,3 +14,129 @@ class PwmInterface
...
+    }
+
+    float GetDutyCycle() override
+    {
+        return static_cast<float>(duty_cycle / match_value);
+    }
+
+    void SetFrequency(uint32_t frequency_hz) override
+    {
+        // disables PWM mode; this will reset all counters to 0
```
kammce: Nit: capitalize the beginnings of your sentences for all comments.

firmware/library/L1_Drivers/pwm.hpp
kammce: Needs to be prefixed with underscore.

delwin-lei:
I have made changes to the SetDutyCycle method. Can you check to see if the changes are valid? It builds and executes properly.

I am also unsure of how to create the constructor for a PinConfigure object properly. I am currently following the example constructor for a PinConfigureInterface & object but am unsure what to initialize my PinConfigure pwm_pin_ to be.

I still have to remove the magic numbers for the PWM register configurations.
@kammce

delwin-lei:
Pwm(const PinConfigureInterface & gpio_):
        pwm_(gpio_),
        pwm_pin_(),
        port_(pwm_pin_.GetPort()),
        channel_(pwm_pin_.GetPin()),
        duty_cycle(0)``

I'm not sure how to initialize pwm_pin_ to the same port and pin number of the PinConfigure object that will be passed in as the argument.

firmware/library/L1_Drivers/pwm.hpp
```diff
@@ -1,13 +1,176 @@
...
+        kActivatePwm1 = (1 << 6)
+    };
+    enum PwmConfigure : uint8_t
+    {
+        kTimerMode = 0,
+        kUsePwmCap0 = (0 << 2),
+        kCounterEnable = (1 << 0),
+        kResetMr0 = (1 << 1),
+        kPwmEnable = (1 << 3)
+    };
```
kammce: The two sections above aren't used by the interface methods. Not only that, but this interface is implementation specific to the LPC40xx boards. These shouldn't be apart of the interface. Remember that an interface is a contract for what ALL pwm classses must follow. If I make a PWM class using this interface for the ESP32, I won't have the same PWM configure register bits.

Please read: http://best-practice-software-engineering.ifs.tuwien.ac.at/patterns/interface.html

firmware/library/L1_Drivers/pwm.hpp
```diff
@@ -1,13 +1,176 @@
...
+        & LPC_PWM1->MR0,
+        & LPC_PWM1->MR1,
+        & LPC_PWM1->MR2,
+        & LPC_PWM1->MR3,
+        & LPC_PWM1->MR4,
+        & LPC_PWM1->MR5,
+        & LPC_PWM1->MR6
+    };
+
+
```
kammce: Remove extra newline.

firmware/library/L1_Drivers/pwm.hpp
```diff
@@ -1,13 +1,176 @@
...
     virtual uint32_t GetFrequency() = 0;
 };
+
+class Pwm : public PwmInterface
+{
+ public:
+    static constexpr uint8_t kPwmFunction = 0b001;
+    static constexpr uint8_t kGpioFunction = 0b000;
+    // The variable that holds the MR values
+    static constexpr volatile uint32_t * match_register_table[7] =
```
kammce: This shouldn't be constexpr, nor defined in the class. This is due to the fact that you need to override this for testing. Put the definition of this in the pwm.cpp file.

firmware/library/L1_Drivers/pwm.hpp
```diff
@@ -1,13 +1,176 @@
...
+    static Pwm CreatePwm()
+    {
+        static_assert(1 <= pwm_port && pwm_port <= 2,
+                      "Available PWM ports are only on port 1 or 2.");
+        static_assert(pwm_pin <= 5,
+                      "Available PWM pins only go up to pin 5.");
+        return Pwm(pwm_port, pwm_pin);
+    }
+
+    constexpr Pwm(uint8_t vpport, uint8_t vppin):
```
kammce: Sorry for the back and forth, but I think it would only be necessary for the user to have to supply a channel, we should only have the user specify the channel and pwm peripheral (I usually call that a port).

Then this class can hold a map to each port/pin for each channel. 
How does that sound?

firmware/library/L1_Drivers/pwm.hpp
```diff
@@ -1,13 +1,176 @@
...
+    {
+        // Sets port to proper function and get pwm channel
+        if (pwm_pin_.GetPort() == 2)
+        {
+            port_ = 1;
+            channel_ = static_cast<uint8_t>(pwm_pin_.GetPin() + 1);
+        }
+    }
+
+    Pwm(const PinConfigureInterface & gpio_):
```
kammce: The underscore is usually meant only for private members of a class, and shouldn't be used in the name of a parameter. Also, a better name would be pin, since a PinConfigure is not a GPIO object.

firmware/library/L1_Drivers/pwm.hpp
```diff
@@ -1,13 +1,176 @@
...
+
+
+    ~Pwm()
+    {
+        pwm_pin_.SetPinFunction(kGpioFunction);
+    }
+
+    void Initialize(uint32_t frequency_hz = 1'000) override
+    {
+        // TODO: Remove when sysclock complete.
```
kammce: This TODO should have an issue number and be directly below the line that needs to be changed.

firmware/library/L1_Drivers/pwm.hpp
```diff
@@ -1,13 +1,176 @@
...
+
+        match_value = config::kSystemClockRate / frequency_hz;
+
+        // Resets PWMTC on Match with MR0
+        LPC_PWM1->MCR |= PwmConfigure::kResetMr0;
+        LPC_PWM1->MR0 = match_value;
+
+        // Enables PWM TC and PC for counting and enables PWM mode
+        LPC_PWM1->TCR = PwmConfigure::kCounterEnable | PwmConfigure::kPwmEnable;
+        LPC_PWM1->CTCR &= PwmConfigure::kTimerMode | PwmConfigure::kUsePwmCap0;
```
kammce: Since both of these configuration variables are zero, this whole register is zero. Did you mean to invert this before you ANDed this?

firmware/library/L1_Drivers/pwm.hpp
```diff
@@ -1,13 +1,176 @@
...
+        // END OF DIRECT MANIPULATION
+
+        match_value = config::kSystemClockRate / frequency_hz;
+
+        // Resets PWMTC on Match with MR0
+        LPC_PWM1->MCR |= PwmConfigure::kResetMr0;
+        LPC_PWM1->MR0 = match_value;
+
+        // Enables PWM TC and PC for counting and enables PWM mode
+        LPC_PWM1->TCR = PwmConfigure::kCounterEnable | PwmConfigure::kPwmEnable;
```
kammce: Can this register be set directly like this? Do you not need to OR this register?

firmware/library/L1_Drivers/pwm.hpp
```diff
@@ -1,13 +1,176 @@
...
+        LPC_PWM1->CTCR &= PwmConfigure::kTimerMode | PwmConfigure::kUsePwmCap0;
+
+
```
kammce: Remove all extra lines.

firmware/library/L1_Drivers/pwm.hpp
```diff
@@ -1,13 +1,176 @@
...
+        // Enables PWM[channel] output
+        LPC_PWM1->PCR |= PwmOutputEnable(channel_);
+
+        pwm_pin_.SetPinFunction(kPwmFunction);
+    }
+
+    void SetDutyCycle(float percentage) override
+    {
+
```
kammce: Remove extra line.

firmware/library/L1_Drivers/pwm.hpp
```diff
@@ -1,13 +1,176 @@
...
+        if (percentage < 0 || percentage > 100)
```
kammce: Prefer to write this, the following way:
`0 < percentage && percentage < 100`

It looks more like `0 < percentage < 100` in math.

firmware/library/L1_Drivers/pwm.hpp
```diff
@@ -1,13 +1,176 @@
...
+        {
+            // TODO: Replace with an assert 
+            return;
+        }
+        duty_cycle = static_cast<uint32_t>((percentage * static_cast<float>(match_value)) / 100);
```
kammce: I think this would be cleaner if this was made into an inline method. You could then test to make sure it works.

firmware/library/L1_Drivers/pwm.hpp
```diff
@@ -1,13 +1,176 @@
...
+
+        // Pointer to point to address of Match Register
+        volatile uint32_t * match_register_selector;
+
+        // Have the pointer point to address of MR
+        match_register_selector = match_register_table[channel_];
+        // Set the value of the MR
+        *match_register_selector = duty_cycle;
```
kammce: I'm not sure why you went to the lengths to make a pointer to the register when you can use it directly. It would look cleaner if you simply use the table. Changing the table from static to constexpr should fix any compiler errors you may have gotten with this.

firmware/library/L1_Drivers/pwm.hpp
```diff
@@ -1,13 +1,176 @@
...
+
+    void SetFrequency(uint32_t frequency_hz) override
+    {
+        // Disables PWM mode; this will reset all counters to 0
+        // And allow us to update MR0
+        LPC_PWM1->TCR &= ~PwmConfigure::kPwmEnable;
+
+        match_value = config::kSystemClockRate / frequency_hz;
+        LPC_PWM1->MR0 = match_value;
+
```
kammce: Remove extra line (I typically consider comments to be newlines.

firmware/library/L1_Drivers/pwm.hpp
```diff
@@ -1,13 +1,176 @@
...
+    float GetDutyCycle() override
+    {
+        return static_cast<float>(duty_cycle / match_value);
+    }
+
+    void SetFrequency(uint32_t frequency_hz) override
+    {
+        // Disables PWM mode; this will reset all counters to 0
+        // And allow us to update MR0
+        LPC_PWM1->TCR &= ~PwmConfigure::kPwmEnable;
```
kammce: Make EnablePwm() a PWM a public interface method. 

It will give users the ability to turn off and off PWM and you can use it to test this method. Notice that you toggle the pwm bit, but there is no way to check that this happened using test by side effect. But you can spy on virtual a class method and see that it was executed with a parameters `EnablePwm(false)` , turns off pwm, and `EnablePwm(true)`, turns it back on in that order.

firmware/library/L1_Drivers/pwm.hpp
```diff
@@ -1,13 +1,176 @@
...
+    {
+        return static_cast<uint8_t>(1 <<  (channel + 8));
+    }
+
+ private:
+    const PinConfigureInterface & pwm_;
+
+    PinConfigure pwm_pin_;
+
+    static uint32_t match_value;
```
kammce: Rather than keeping this local to the class, you could use a constexpr * match_value. This is not thread safe. The source of truth about the match value, should be the register not a variable.

firmware/library/L1_Drivers/pwm.hpp
```diff
@@ -1,13 +1,164 @@
...
+
+        // Enables PWM[channel] output
+        pwm1->PCR |= PwmOutputEnable(channel_);
+
+        pwm_pin_.SetPinFunction(kPwmFunction);
+    }
+
+    void SetDutyCycle(float percentage) override
+    {
+        SJ2_ASSERT_FATAL(0 < percentage && percentage < 100,
```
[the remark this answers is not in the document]
delwin-lei: Did you mean 100.0 or scale down to 1.0 as 100%?

firmware/library/L1_Drivers/pwm.hpp
```diff
@@ -1,13 +1,164 @@
...
+
+    void Initialize(uint32_t frequency_hz = 1'000) override
+    {
+        // DIRECT MANIUPULATION OF SYSTEM CLOCK REGISTERS
+        // WILL REPLACE WITH FUNCTIONS LATER
+        // Enables PWM1 power/clock control bit
+        sc->PCONP |= SysClk::kActivatePwm1;
+        // TODO(delwin-lei): Remove when sysclock complete.
+        // END OF DIRECT MANIPULATION
+
```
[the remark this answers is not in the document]
delwin-lei: Why do we put it in Initialize rather than the constructor?

firmware/library/L1_Drivers/pwm.hpp
```diff
@@ -1,13 +1,164 @@
...
+        // Resets PWMTC on Match with MR0
+        pwm1->MCR |= PwmConfigure::kResetMr0;
+        pwm1->MR0 = config::kSystemClockRate / frequency_hz;;
+
+        // Enables PWM TC and PC for counting and enables PWM mode
+        EnablePwm();
+        pwm1->CTCR &= ~PwmConfigure::kTimerMode;
```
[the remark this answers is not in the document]
delwin-lei: Not sure what else to call it. Any suggestions?

firmware/library/L1_Drivers/pwm.hpp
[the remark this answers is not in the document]
kammce: I would rather put it in the initialize, so we know that this assert is happening after the application has started. The constructor happens pretty early in the system initialization. For now, put it in the init function.

firmware/library/L1_Drivers/pwm.hpp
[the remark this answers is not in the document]
kammce: What does it mean?

firmware/library/L1_Drivers/pwm.hpp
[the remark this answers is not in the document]
kammce: 1.0 = 100%.

firmware/library/L1_Drivers/pwm.hpp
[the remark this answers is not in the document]
kammce: Ah, ok, so the mode is, "timer". This is fine then.

firmware/library/L1_Drivers/pwm.hpp
```diff
@@ -1,13 +1,168 @@
...
+    virtual void EnablePwm(bool enable = true) = 0;
+    virtual uint32_t GetMatchRegister0() = 0;
+};
+
+class Pwm : public PwmInterface
+{
+ public:
+    static constexpr uint8_t kPwmFunction = 0b001;
+    static constexpr uint8_t kGpioFunction = 0b000;
+    // The variable that holds the MR values
```
kammce: Remove this line. Quite obvious given the name of the variable.

firmware/library/L1_Drivers/pwm.hpp
```diff
@@ -1,13 +1,168 @@
...
+
+    ~Pwm()
+    {
+        pwm_pin_.SetPinFunction(kGpioFunction);
+    }
+
+    void Initialize(uint32_t frequency_hz = 1'000) override
+    {
+        SJ2_ASSERT_FATAL(channel_ < 6,
+         "Channel selected is out of bounds.");
```
kammce: Add, "channel must be between 0 to 5.

firmware/library/L1_Drivers/pwm.hpp
```diff
@@ -1,13 +1,168 @@
...
+
+    void EnablePwm(bool enable = true) override
+    {
+        if(enable)
+        {
+            pwm1->TCR |= PwmConfigure::kCounterEnable | PwmConfigure::kPwmEnable;
+        }
+        else
+        {
+            pwm1->TCR &= ~PwmConfigure::kPwmEnable;
```
kammce: Why don't you clear the counter enable as well? Or is it not necessary if you simply disable the whole pwm module.

firmware/library/L1_Drivers/pwm.hpp
```diff
@@ -1,13 +1,168 @@
...
+    {
+        return (1 << (channel + 8));
+    }
+
+ private:
+    const PinConfigureInterface & pwm_;
+
+    PinConfigure pwm_pin_;
+
+    // Will signify current PWM0/1 and channel
```
kammce: Remove the 0/ from PWM0/1

firmware/library/L1_Drivers/pwm_test.cpp
```diff
@@ -0,0 +1,94 @@
...
+
+    Fake(Method(mock_pwm_pin, SetPinFunction));
+    PinConfigureInterface &pwm = mock_pwm_pin.get();
+    Pwm test2_0(pwm);
+
+    constexpr uint32_t kDefaultFrequency = 1'000;
+    constexpr uint8_t kActivePwm = (1 << 6);
+    constexpr uint8_t kResetMr0 = (1 << 1);
+    constexpr uint8_t kCounterEnable = (1 << 0);
+    constexpr uint8_t kTimerMode = (0b11 << 0);
```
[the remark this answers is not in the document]
kammce: Ignore this.

firmware/library/L1_Drivers/pwm.hpp
```diff
@@ -1,13 +1,169 @@
...
+
+        // Enables PWM[channel] output
+        pwm1->PCR |= PwmOutputEnable(channel_);
+
+        pwm_pin_.SetPinFunction(kPwmFunction);
+    }
+
+    void SetDutyCycle(float percentage) override
+    {
+        SJ2_ASSERT_FATAL(0.0 <= percentage/100.0 && percentage/100.0 <= 1.0,
```
kammce: This should only take a value from 0.0 to 1.0. You do not need to do these calculations in the assert fatal. You may want to change the variable name back to duty cycle and add a comment on the range that is supposed to be used for this.

firmware/library/L1_Drivers/pwm.hpp
```diff
@@ -1,13 +1,169 @@
...
+            "Percentage of Duty Cycle provided is out of bounds.");
+        EnablePwm(false);
```
kammce: Are you sure that you need to disable PWM each time you change the duty cycle?

delwin-lei: Maybe. When I didn't have the EnablePwm function it would stop working when I changed the duty cycle until I set the frequency. This situation happens when I go to a smaller duty cycle than the previous value. I think I can achieve the same result by just resetting the counter but I won't be able to confirm until I test it on an oscilloscope. EnablePwm and resetting the counter are located in the same register. The bit that disables and enables the pwm automatically resets the counter which was why I chose to go with enabling and disabling the pwm.

firmware/library/L1_Drivers/pwm.hpp
delwin-lei: So the input to the SetDutyCycle function should be a decimal?
e.g 0.5 = 50%

PR #153, opened 2018-08-07
Temperature Sensor Interface

ChuefengVang:


firmware/library/L4_Application/Temperature_Sensor.hpp
```diff
@@ -0,0 +1,21 @@
+/* Interface used to obtain the temperature of the system.*/
```
kammce: Use // for all comments. No multiline comments.

firmware/library/L4_Application/Temperature_Sensor.hpp
```diff
@@ -0,0 +1,21 @@
...
+#pragma once
+
+class TemperatureInterface
+
```
kammce: Remove extra line.

firmware/library/L4_Application/Temperature_Sensor.hpp
```diff
@@ -0,0 +1,21 @@
...
+{
+ public:
+  virtual bool Init() = 0;
+  // Initialize Drivers
```
kammce: Um, comments are usually, and most intuitively placed above the line it talks about. But regardless, the comment is superfluous. The name of the method already tells you what it does.

firmware/library/L4_Application/Temperature_Sensor.hpp
```diff
@@ -0,0 +1,21 @@
+/* Interface used to obtain the temperature of the system.*/
+#pragma once
+
+class TemperatureInterface
+
+{
+ public:
+  virtual bool Init() = 0;
```
kammce: Spell out the method names fully. Do not short hand.

firmware/library/L4_Application/Temperature_Sensor.hpp
```diff
@@ -0,0 +1,21 @@
...
+  // Initialize Drivers
+
+  virtual unsigned char
+  GetTemperatureByte() = 0;
```
kammce: I'm not sure why this is on a newline. Also use uint8_t, rather than unsigned char. You cannot guarantee the size of unsigned char.

ChuefengVang: For some reason lint does not like it. It kept telling me that each line should have at most 80 characters.

kammce: If that's the case, then you need to make sure no lines exceed 80 characters. You can use new lines to do this. You can also run clang-format on your source file to reformat it.

PR #155, opened 2018-08-08
Temperature Sensor Interface

ChuefengVang:


firmware/library/L4_Application/Temperature_Sensor.hpp
```diff
@@ -0,0 +1,17 @@
+// Interface used to obtain the temperature of the system.
```
kammce: More like, sense the temperature of the temperature sensor.

ChuefengVang: Got it, I'll update that.

kammce:
You cannot be merged in because your branch is out of date. Please update your branch.

PR #156, opened 2018-08-08
Adding GPIO example code

PricelessAntonio:


firmware/examples/Gpio/source/main.cpp
```diff
@@ -0,0 +1,33 @@
+#include "L1_Drivers/gpio.hpp"
+
```
kammce: You are missing the debug_print.hpp include. Be sure to include it.

firmware/examples/Gpio/source/main.cpp
```diff
@@ -0,0 +1,33 @@
...
+int main(void)
+{
+    // Port 2, pin 0 must be connected to a switch that is able to ground and
```
kammce: Can you use one of the pins on the SJTwo board that is connected to a Button and LED? Then put in a note, saying that pressing the onboard button should light up the led.

firmware/examples/Gpio/source/main.cpp
```diff
@@ -0,0 +1,33 @@
...
+    // supply 3.3 volts to the pin (You can use a 1kOhm resistor to connect the
+    // pin to the switch for protection).
+    // Port 2, pin 2 must be connected to an LED that connects to ground (use
+    // the appropriate resistor to protect the LED).
```
kammce: You may also want to link to an image of what this looks like.

firmware/examples/Gpio/source/main.cpp
```diff
@@ -0,0 +1,33 @@
...
+    Gpio p2_00(2, 0);
+    Gpio p2_02(2, 2);
+
+    DEBUG_PRINT("Configure port 2, pin 0 as input and port 2, pin 2 as output");
```
kammce: If you see the other examples. They typically start with, "Starting ____ application..." Can you follow that scheme.
Each debug print should be its own thing, so you before you set P2.00 to input, you shoudl have a message and another for setting p2.02 to output.

firmware/examples/Gpio/source/main.cpp
PricelessAntonio: If I choose to use the on-board SJTwo led and button, would I still need to link an image?

firmware/examples/Gpio/source/main.cpp
PricelessAntonio: Is there a reason why the other examples don't include the debug_print.hpp file when DEBUG_PRINT() is used?

kammce: Oh, if that is the case, then it needs to include that. It may been included through another include, but it should be included.

kammce:
Thank you @PricelessAntonio.
Merging this pull request resolves #7.

PR #157, opened 2018-08-09
Dac Implementation

RadConMannBor:
Pull request necessary for code review,  I know i need many changes and i currently need to figure out the correct calculation to implement the frequency into all my demo functions.

### DIFF
--- a/firmware/examples/Dac/source/main.cpp
+++ b/firmware/examples/Dac/source/main.cpp
@@ -1,0 +1,141 @@
+// Digital-to-Analog Demoonstration Main, Designed so that
+// The DAC driver can be demonstrated on an Oscilloscope.
+#include "L0_LowLevel/delay.hpp"
+#include "L0_LowLevel/LPC40xx.h"
+#include "L1_Drivers/dac.hpp"
+#include "L2_Utilities/debug_print.hpp"
+// Input Number for starting demo
+
+const uint16_t kSineLookup[256] =
+{
+    500, 512, 525, 537, 549, 561, 573, 585,
+    598, 610, 621, 633, 645, 657, 668, 680,
+    691, 703, 714, 725, 736, 746, 757, 767,
+    778, 788, 798, 808, 817, 827, 836, 845,
+    854, 862, 870, 879, 887, 894, 902, 909,
+    916, 922, 929, 935, 941, 947, 952, 957,
+    962, 966, 971, 975, 978, 982, 985, 988,
+    990, 993, 995, 996, 998, 999, 999, 1000,
+    1000, 1000, 999, 999, 998, 996, 995, 993,
+    990, 988, 985, 982, 978, 975, 971, 966,
+    962, 957, 952, 947, 941, 935, 929, 922,
+    916, 909, 902, 894, 887, 879, 870, 862,
+    854, 845, 836, 827, 817, 808, 798, 788,
+    778, 767, 757, 746, 736, 725, 714, 703,
+    691, 680, 668, 657, 645, 633, 621, 610,
+    598, 585, 573, 561, 549, 537, 525, 512,
+    500, 488, 475, 463, 451, 439, 427, 415,
+    402, 390, 379, 367, 355, 343, 332, 320,
+    309, 297, 286, 275, 264, 254, 243, 233,
+    222, 212, 202, 192, 183, 173, 164, 155,
+    146, 138, 130, 121, 113, 106, 98, 91,
+    84, 78, 71, 65, 59, 53, 48, 43,
+    38, 34, 29, 25, 22, 18, 15, 12,
+    10, 7, 5, 4, 2, 1, 1, 0,
+    0, 0, 1, 1, 2, 4, 5, 7,
+    10, 12, 15, 18, 22, 25, 29, 34,
+    38, 43, 48, 53, 59, 65, 71, 78,
+    84, 91, 98, 106, 113, 121, 130, 138,
+    146, 155, 164, 173, 183, 192, 202, 212,
+    222, 233, 243, 254, 264, 275, 286, 297,
+    309, 320, 332, 343, 355, 367, 379, 390,
+    402, 415, 427, 439, 451, 463, 475, 488
+};
+
+void SineDac(Dac * dac_demo, uint32_t delay)
+{
+    SJ2_ASSERT_FATAL(delay < 500,
+                     "Input frequency out of range! Frequency "
+                     "input needs to be between 0 and 500 Hz.");
+    for ( uint8_t i = 0; i < 255; i++ )
+    {
+        dac_demo->WriteDac(kSineLookup[i]);
+        Delay(delay);
+    }
+}
+void SawtoothDac(Dac * dac_demo, uint32_t delay)
+{
+    SJ2_ASSERT_FATAL(delay < 500,
+                     "Input delay out of range! Delay "
+                     "input needs to be between 0 and 500 Hz.");
+    int sawtooth = 0;
+    for ( uint8_t i = 0; i < 255; i++ )
+    {
+        uint16_t conversion = static_cast<uint16_t>(sawtooth);
+        dac_demo->WriteDac(conversion);
+        sawtooth = sawtooth + 4;
+        Delay(delay);
+    }
+}
+void TriangleDac(Dac * dac_demo, uint32_t delay)
+{
+    SJ2_ASSERT_FATAL(delay < 500,
+                     "Input delay out of range! Delay "
+                     "input needs to be between 0 and 500 Hz.");
+    int triangle = 0;
+    for ( uint8_t i = 0; i < 127; i++ )
+    {
+        uint16_t conversion = static_cast<uint16_t>(triangle);
+        dac_demo->WriteDac(conversion);
+        triangle = triangle + 8;
+        Delay(delay);
+    }
+    for ( uint8_t i = 127; i < 255 ; i++ )
+    {
+        uint16_t conversion = static_cast<uint16_t>(triangle);
+        dac_demo->WriteDac(conversion);
+        triangle = triangle - 8;
+        Delay(delay);
+    }
+}
+void SerratedDac(Dac * dac_demo, uint32_t delay)
+{
+    SJ2_ASSERT_FATAL(delay < 500,
+                     "Input delay out of range! Delay "
+                     "input needs to be between 0 and 500.");
+    int serrated = 0;
+    for ( uint8_t i = 0; i < 127; i++ )
+    {
+        uint16_t conversion = static_cast<uint16_t>(serrated);
+        dac_demo->WriteDac(conversion);
+        serrated = serrated + 8;
+        Delay(delay);
+    }
+    serrated = 0;
+    for ( uint8_t i = 127; i < 253 ; i++ )
+    {
+        uint16_t conversion = static_cast<uint16_t>(serrated);
+        dac_demo->WriteDac(conversion);
+        serrated = serrated + 4;
+        Delay(delay);
+    }
+}
+void (*dac_controllers[])(Dac *, uint32_t) = {
+     SerratedDac,
+     TriangleDac,
+     SawtoothDac,
+     SineDac
+};
+
+void StartDemo(Dac dac_demo, uint32_t input_cycles)
+{
+    for (uint32_t i = 0; i < 4 ; i++)
+    {
+        for (int j = 0; j < 30; j++)
+        {
+            dac_controllers[i](&dac_demo, input_cycles);
+        }
+    }
+}
+int main(void)
+{
+    Dac dac;
+    dac.Initialize();
+    DEBUG_PRINT(
+        "Hook up pin p0.26 to an oscilloscope to test if it works!\n");
+    DEBUG_PRINT("Starting Output of waves...\n");
+    while (1)
+    {
+        StartDemo(dac, 5);
+    }
+}
--- a/firmware/library/L1_Drivers/dac.cpp
+++ b/firmware/library/L1_Drivers/dac.cpp
@@ -1,0 +1,5 @@
+#include "L1_Drivers/dac.hpp"
+
+LPC_DAC_TypeDef * Dac::dac_register = LPC_DAC;
+LPC_SC_TypeDef * Dac::sc_ptr = LPC_SC;
+LPC_IOCON_TypeDef * Dac::iocon_register = LPC_IOCON;
--- a/firmware/library/L1_Drivers/dac.hpp
+++ b/firmware/library/L1_Drivers/dac.hpp
@@ -1,19 +1,97 @@
+// Dac is a driver for the Digital-To-Analog converter for the LPC40xx chip
 #pragma once
-
-#include <cstdint>
-#include <cstdio>
+#include "L0_LowLevel/LPC40xx.h"
+#include "L1_Drivers/pin_configure.hpp"
 
 class DacInterface
 {
  public:
-    virtual bool Initialize();
-    virtual bool WriteDac(uint16_t dac_output);
-    virtual bool SetVoltage(float voltage);
-    // Sine Demonstration Function
-    virtual void SineDac(uint32_t frequency_hz);
-    // Sawtooth Demonstration Function
-    virtual void SawtoothDac(uint32_t frequency_hz);
-    // Triangle Demonstration Function
-    virtual void TriangleDac(uint32_t frequency_hz);
-    virtual void SetBias(bool bias_level);
+    virtual void Initialize(void)                   = 0;
+    virtual bool WriteDac(uint16_t dac_output)      = 0;
+    virtual bool SetVoltage(float voltage)          = 0;
+};
+class Dac : public DacInterface
+{
+ public:
+    static LPC_DAC_TypeDef * dac_register;
+    static LPC_SC_TypeDef * sc_ptr;
+    static LPC_IOCON_TypeDef * iocon_register;
+    static constexpr float    kVref = 3.3f;
+    static constexpr uint8_t  kDacMode = 0b010;
+    static constexpr uint16_t kClearDac = 0b1111111111;
+    static constexpr uint8_t  kClockDiv = 0b0010;
+    enum Bit : uint8_t
+    {
+        kClockDividerReg = 1,
+        kBiasReg         = 16,
+        kDacOutReg       = 6,
+        kEnableDacReg    = 16,
+        kPullUpDown      = 3,
+        kFunc            = 0,
+        kPort0           = 0,
+        kPin26           = 26
+    };
+    enum class BiasLevel: uint8_t
+    {
+        kBiasHigh  = 0,
+        kBiasLow   = 1
+    };
+     constexpr Dac() :
+                        dac_(&dac_pin_),
+                        dac_pin_(kPort0, kPin26)
+     {
+     }
+     // For unit testing mocking purposes
+     explicit constexpr Dac(PinConfigureInterface * dac_pin) :
+                      dac_(dac_pin),
+                      dac_pin_(PinConfigure::CreateInactivePin())   // P0.26
+     {
+     }
+    // Initialize Dac by setting the clock divider and enabling
+    // The Dac Pin withiin IOCON.  Then set the initial Bias level to 0
+    void Initialize(void) override
+    {
+        dac_->SetPinFunction(kDacMode);
+        dac_->EnableDac(true);
+        dac_->SetAsAnalogMode(true);
+        dac_->SetPinMode(PinConfigureInterface::kInactive);
+        // Set Update Rate to 1MHz
+        SetBias(BiasLevel::kBiasHigh);
+    }
+    // The Dac ouptut is a 10 bit input and thus it is necessary to
+    // ensure dac_output is less than 1024(largest 10-bit number)
+    bool WriteDac(uint16_t dac_output) override
+    {
+        SJ2_ASSERT_FATAL(dac_output < 1023,
+                         "dac_output is above SJ2 output capacity."
+                         "output capacity must be between 0-1023.");
+        dac_register->CR &= ~(kClearDac << kDacOutReg);
+        dac_register->CR |= (dac_output << kDacOutReg);
+        return true;
+    }
+    // Takes an input voltage and converts the float value and calculates
+    // the conversion necessary and then typecasts it to an integer.
+    // If the voltage value is greater than 3.3 it will fail and end.
+    bool SetVoltage(float voltage) override
+    {
+        // value = (dac_out * 1024)/VrefP
+        float value = (voltage * 1024.0f) / kVref;
+        uint16_t   conversion = static_cast<uint16_t>(value);
+        SJ2_ASSERT_FATAL(voltage < kVref,
+                         "Input voltage above SJ2"
+                         "output capacity of 3.3 Volts.");
+        WriteDac(conversion);
+        return true;
+    }
+    // Sets the Bias for the Dac, which determines the settling time
+    // , max current, and the allowed maximum update rate
+    void SetBias(BiasLevel bias_level)
+    {
+        uint8_t bias = static_cast<int>(bias_level);
+        dac_register->CR = (dac_register->CR &~(1 << kBiasReg))
+                            |(bias << kBiasReg);
+    }
+ private:
+    PinConfigureInterface * dac_;
+    PinConfigure dac_pin_;
 };
--- a/firmware/library/L1_Drivers/dac_test.cpp
+++ b/firmware/library/L1_Drivers/dac_test.cpp
@@ -1,0 +1,89 @@
+#include "L1_Drivers/dac.hpp"
+#include "L1_Drivers/pin_configure.hpp"
+#include "L5_Testing/testing_frameworks.hpp"
+TEST_CASE("Testing Dac", "[dac]")
+{
+    using fakeit::Fake;
+    using fakeit::Mock;
+    using fakeit::Spy;
+    using fakeit::Verify;
+    using fakeit::When;
+    using fakeit::Using;
+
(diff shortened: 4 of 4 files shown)


firmware/library/L1_Drivers/dac.hpp
```diff
@@ -1,19 +1,209 @@
...
+        EnableDAC(true);
+        // Set Update Rate to 1MHz
+        LPC_DAC->CR &= ~(1 << kBiasReg);
+        // return true;
+    }
+    // The Dac ouptut is a 10 bit input and thus it is necessary to
+    // ensure dac_output is less than 1024(largest 10-bit number)
+    bool WriteDac(uint16_t dac_output) override
+    {
+        if ( dac_output < 1024 )
```
kammce: Rather than doing this, use an SJ2_ASSERT_FATAL() for this. See i2c.hpp for an example.

firmware/library/L1_Drivers/dac.hpp
```diff
@@ -1,19 +1,209 @@
...
+    }
+    // Takes an input voltage and converts the float value and calculates
+    // the conversion necessary and then typecasts it to an integer.
+    // if the voltage value is greater than 3.3 it will fail
+    bool SetVoltage(float voltage) override
+    {
+        // value = (dac_out * 1024)/VrefP
+        float value = (voltage * 1024) / kVrefP;
+        int   conversion = static_cast<int>(value);
+        if ( voltage < 3.3 )
```
kammce: another case for ASSERT.

firmware/library/L1_Drivers/dac.hpp
```diff
@@ -1,19 +1,209 @@
...
+    // Demonstration Function for Triangle demo,  Still need to determine
+    // the correct way to incorperate frequency into the output
+    void TriangleDac(uint32_t frequency_hz) override
+    {
+        for ( uint8_t i = 0; i < 255; i++ )
+        {
+            LPC_DAC->CR = (kTriangleLookup[i] << kDacOutReg);
+            // TODO(Conrad) Determine wait time based on frequency and testing
+        }
+    }
```
kammce: I think these demonstration methods should only be apart of the example project and not the class.

RadConMannBor: Ok I have moved them to my main for example project.  Should the main be code reviewed as well?  Its not going to be pulled into the repository is it?

RadConMannBor:
My code builds and compiles locally but on make presubmit and travic CI it says build failed for some reason.  I added in the SJ2_ASSERT_FATAL instead of my if statements and they work fine with make build, but not make presubmit.

firmware/library/L1_Drivers/dac.hpp
```diff
@@ -1,19 +1,76 @@
...
-    // Sawtooth Demonstration Function
-    virtual void SawtoothDac(uint32_t frequency_hz);
-    // Triangle Demonstration Function
-    virtual void TriangleDac(uint32_t frequency_hz);
-    virtual void SetBias(bool bias_level);
+    enum BiasLevel: uint8_t
+    {
+        kBiasHigh  = 0,
+        kBiasLow   = 1
+    };
```
kammce: I still think you should move this to the Dac class, since it is specific to the LPC4078. Also, make this into a `enum class`.

firmware/library/L1_Drivers/dac.hpp
```diff
@@ -1,19 +1,76 @@
...
+    enum Bit : uint8_t
+    {
+        kClockDividerReg = 0,
+        kBiasReg = 16,
+        kDacOutReg = 6,
+        kEnableDacReg = 16
+    };
+    // PinConfigure led3(PinConfigure::CreatePinConfigure<1, 15>());
+    /* Initialize Dac by setting the clock divider and enabling 
+      The Dac Pin withiin IOCON.  Then set the initial Bias level to 0 */
```
kammce: Move to using // for comments.

firmware/library/L1_Drivers/dac.hpp
```diff
@@ -1,19 +1,76 @@
...
+    static LPC_IOCON_TypeDef * iocon_ptr;
+    static constexpr float kVref = 3.3;
+    enum Bit : uint8_t
+    {
+        kClockDividerReg = 0,
+        kBiasReg = 16,
+        kDacOutReg = 6,
+        kEnableDacReg = 16
+    };
+    // PinConfigure led3(PinConfigure::CreatePinConfigure<1, 15>());
```
kammce: Uncomment this line, put it in private and use it for this class.

firmware/library/L1_Drivers/dac.hpp
```diff
@@ -1,19 +1,76 @@
...
+    /* Initialize Dac by setting the clock divider and enabling 
+      The Dac Pin withiin IOCON.  Then set the initial Bias level to 0 */
+    void Initialize(void) override
+    {
+        // Set Clock Divider to 1
+        sc_ptr->PCLKSEL = (1 << kClockDividerReg);
```
kammce: This is massive problem. This will result in you disabling all of the clocks except for your DAC clock. You need to OR this.

firmware/library/L1_Drivers/dac.hpp
```diff
@@ -1,19 +1,76 @@
...
+        // Enable DAC Pin P0.26
+        iocon_ptr->P0_26 = (1 << kEnableDacReg);
```
kammce: Use the pin configure class, do not use the registers directly. Try to stay away modifying registers as much as possible.

firmware/library/L1_Drivers/dac.hpp
```diff
@@ -1,19 +1,76 @@
...
+        // Set Update Rate to 1MHz
+        dac_ptr->CR &= ~(1 << kBiasReg);
+    }
+    // The Dac ouptut is a 10 bit input and thus it is necessary to
+    // ensure dac_output is less than 1024(largest 10-bit number)
+    bool WriteDac(uint16_t dac_output) override
+    {
+        SJ2_ASSERT_FATAL(dac_output > 1023,
+                         "dac_output is above SJ2"
+                         "output capacity.");
```
kammce: You can be more specific here. For example you can say, `Input invalid. Input MUST be between 0 and 1023`. So that the user immediately knows what they need to fix. Otherwise, the developer will have to look up what the SJ2's output capacity is.

firmware/library/L1_Drivers/dac.hpp
```diff
@@ -1,19 +1,76 @@
...
+    // the conversion necessary and then typecasts it to an integer.
+    // If the voltage value is greater than 3.3 it will fail and end.
+    bool SetVoltage(float voltage) override
+    {
+        // value = (dac_out * 1024)/VrefP
+        float value = (voltage * 1024) / kVref;
+        int   conversion = static_cast<int>(value);
+        SJ2_ASSERT_FATAL(voltage > kVref,
+                         "Input voltage above SJ2"
+                         "output capacity of 3.3 Volts.");
```
kammce: This has a good amount of information and context!

firmware/library/L1_Drivers/dac.hpp
```diff
@@ -1,19 +1,76 @@
...
+        dac_ptr->CR = (conversion << kDacOutReg);
+        return true;
+    }
+    // Sets the Bias for the Dac, which determines the settling time
+    // , max current, and the allowed maximum update rate
+    void SetBias(BiasLevel bias_level) override
```
kammce: If you don't make BiasLevel an `enum class`, the user can put in any integer value they like for this, and since you are not using an AND to remove any of the other bits, this can result in additional bits being manipulated. Its better to add that `class` keyword to prevent the user from getting to that point.

firmware/library/L1_Drivers/dac.hpp
RadConMannBor: I use it as a parameter to pass into SetBias(BiasLevel bias_level); 
Thus I think I need it to be in the interface.

firmware/library/L1_Drivers/dac.hpp
RadConMannBor: I am honestly extremely confused on how to use the pin configure class,  Does pin configure somehow point to the register? such as LPC_IOCON or does it do it in some other way?  Does it not modify the registers?  I have tried looking at I2C and GPIO for it but is there any other ones you suggest for me to understand it better? Like  For LPC_SC->PCLKSEL is this even correlated with a pin and port?  Isn't it a specific register that it is pointed at?

firmware/library/L1_Drivers/dac.hpp
kammce: Ah, I see. Then remove SetBias from the interface then. It should only be in the implementation of the driver.

kammce:
When you get a hard fault, I setup the system to drop the PC address of the caller that caused the hard fault to occur. You can use that with `arm-none-eabi-addr2line` to find which line of code caused this.

firmware/library/L1_Drivers/dac.hpp
kammce: > Does pin configure somehow point to the register?
Yes, just like how your DAC class points to the DAC registers.
> such as LPC_IOCON or does it do it in some other way? 
Nope, it uses LPC_IOCON. Just does all of the work for you. It is driver to handle all of that work.
> Does it not modify the registers?
It does. But your driver shouldn't modify the LPC_IOCON register. Why waste time doing that, when there is already a drive that does that for you.
>  I have tried looking at I2C and GPIO for it but is there any other ones you suggest for me to understand it better?
Lets think about PinConfigure as if it were a GPIO. If you had a GPIO object in your class, and you gave it a specific port and pin, you should be able to set it as output and pull it high or low, or set it as an input and read if it is high or low. Going back to PinConfigure, does the same thing in nature, in that it configures pins. so you can set them as pull up, pull down, repeater mode. You can set the pin as DAC function, or GPIO function or a PWM, all depending on the pin in question. Its a driver that configures pins. Look at the methods and see what you can tell it to do.
>  Like For LPC_SC->PCLKSEL is this even correlated with a pin and port?
Nope, that has to do with clocks.
>  Isn't it a specific register that it is pointed at?
PinConfigure points to LPC_IOCON, but only during an actual build. During unit tests, it is pointed to a local storage area, where we can read the side effects of the driver's function calls.

firmware/library/L1_Drivers/dac.cpp
```diff
@@ -0,0 +1,8 @@
+#include "L1_Drivers/dac.hpp"
+
+// Register Pointer for LPC_DAC
```
kammce: This comment is not necessary. Its self explanatory.

firmware/library/L1_Drivers/dac.cpp
```diff
@@ -0,0 +1,8 @@
...
+LPC_DAC_TypeDef * Dac::dac_ptr = LPC_DAC;
+// Register Pointer for LPC_SC (System Clock)
```
kammce: This comment is not necessary. Its self explanatory.

firmware/library/L1_Drivers/dac.cpp
```diff
@@ -0,0 +1,8 @@
...
+LPC_SC_TypeDef * Dac::sc_ptr = LPC_SC;
+// Register Pointer for LPC_IOCON (Pin Connections)
```
kammce: This comment is not necessary. Its self explanatory.

firmware/examples/Dac/source/main.cpp
```diff
@@ -0,0 +1,160 @@
...
+    Dac * dac = new Dac();
+    dac->Initialize();
+    fputs("Enter anything to start demo: \n", stdout);
+    scanf("%lu", &start);
+    DEBUG_PRINT(
+        "Hook up pin p0.26 to an oscilloscope to test if it works\n");
+    while (1)
+    {
+        fputs("Enter a delay for Waveforms: \n", stdout);
+        scanf("%lu", &cycles);
```
kammce: I would prefer if the demos just ran without needing input from the user.

firmware/examples/Dac/source/main.cpp
```diff
@@ -0,0 +1,160 @@
...
+    // Clears the terminal
+    fputs("\e[2J\e[H", stdout);
+    #endif
+    fputs(SJ2_BACKGROUND_GREEN
+         "================================== SJTwo Booted! "
+         "==================================\n" SJ2_COLOR_RESET, stdout);
+    Dac * dac = new Dac();
+    dac->Initialize();
+    fputs("Enter anything to start demo: \n", stdout);
+    scanf("%lu", &start);
```
kammce: I would prefer if the demos just ran without needing input from the user.

firmware/examples/Dac/source/main.cpp
```diff
@@ -0,0 +1,160 @@
...
+int main(void)
+{
+    #if SJ2_ENABLE_ANSI_CODES
+    // Clears the terminal
+    fputs("\e[2J\e[H", stdout);
+    #endif
+    fputs(SJ2_BACKGROUND_GREEN
+         "================================== SJTwo Booted! "
+         "==================================\n" SJ2_COLOR_RESET, stdout);
+    Dac * dac = new Dac();
```
kammce: Even though this is not apart of the library, it is apart of the demos, which are apart of the library, which means do not allocate memory dynamically. Replace this with:

    Dac dac;

firmware/examples/Dac/source/main.cpp
```diff
@@ -0,0 +1,160 @@
...
+            SawtoothDac(dac_demo, input_cycles);
+        }
+        for (int j = 0; j < 60; j++)
+        {
+            TriangleDac(dac_demo, input_cycles);
+        }
+        for (int j = 0; j < 60; j++)
+        {
+            SerratedDac(dac_demo, input_cycles);
+        }
```
kammce: A lookup table of functions would make this more concise.

firmware/examples/Dac/source/main.cpp
```diff
@@ -0,0 +1,160 @@
...
+    serrated = 0;
+    for ( uint8_t i = 127; i < 253 ; i++ )
+    {
+        uint16_t conversion = static_cast<uint16_t>(serrated);
+        dac_demo->WriteDac(conversion);
+        serrated = serrated + 4;
+        Delay(delay);
+    }
+}
+void StartDemo(Dac * dac_demo, uint32_t input_cycles)
```
kammce: A lookup table of functions would make this more concise.
I think the contents, after you make this change can be placed within the main function.

firmware/examples/Dac/source/main.cpp
```diff
@@ -0,0 +1,160 @@
...
+// The DAC driver can be demonstrated on an Oscilloscope.
+#include "L0_LowLevel/delay.hpp"
+#include "L0_LowLevel/LPC40xx.h"
+#include "L1_Drivers/dac.hpp"
+#include "L2_Utilities/debug_print.hpp"
+// Input Number for starting demo
+volatile uint32_t number = 0;
+volatile uint32_t cycles = 500'000;
+volatile uint32_t start = 500'000;
+volatile int saw = 4;
```
kammce: This variable is unused. If its not needed, it should be removed.

kammce: If you are not using any of these variables, remove them.

firmware/examples/Dac/source/main.cpp
RadConMannBor: Im a little confused what you mean by this?  The lookup table would be just all 4 function calls? and a for loop would call them the same way?

firmware/library/L1_Drivers/dac_test.cpp
```diff
@@ -0,0 +1,110 @@
+// Test for Dac class.
+// Using a test by side effect on the LPC_IOCON register
```
kammce: You can remove the above comments

firmware/library/L1_Drivers/dac_test.cpp
```diff
@@ -0,0 +1,110 @@
...
+    using fakeit::Using;
+
+    LPC_IOCON_TypeDef local_iocon;
+    memset(&local_iocon, 0, sizeof(local_iocon));
+    LPC_DAC_TypeDef local_dac_port;
+    Dac::dac_register = &local_dac_port;
+    PinConfigure iocon_pin_config(0, 26);
+    PinConfigure::pin_map =
+        reinterpret_cast<PinConfigure::PinMap_t *>(&local_iocon);
+    PinConfigure test_pin(0, 26);
```
kammce: Pin configure shouldn't be tested via side effect. It should be tested soley though Mock using fakeit::Mock.

firmware/library/L1_Drivers/dac_test.cpp
```diff
@@ -0,0 +1,110 @@
...
+    Mock<PinConfigureInterface> mock_dac;
```
kammce: Should rename this to mock_dac_pin, rather than mock_dac, which looks like you are mocking your whole library under test.

firmware/examples/Dac/source/main.cpp
RadConMannBor: Im a little confused by what you mean?  Would the lookup table have the function calls?

kammce: Yes

firmware/examples/Dac/dac.hpp
```diff
@@ -0,0 +1,99 @@
...
+        dac_register->CR |= (dac_output << kDacOutReg);
+        return true;
+    }
+    // Takes an input voltage and converts the float value and calculates
+    // the conversion necessary and then typecasts it to an integer.
+    // If the voltage value is greater than 3.3 it will fail and end.
+    bool SetVoltage(float voltage) override
+    {
+        // value = (dac_out * 1024)/VrefP
+        float value = (voltage * 1024) / kVref;
```
kammce: make 1024 -> `1024.0f` to force the compiler to use it as a float.

firmware/examples/Dac/dac.hpp
```diff
@@ -0,0 +1,99 @@
...
+    // Initialize Dac by setting the clock divider and enabling
+    // The Dac Pin withiin IOCON.  Then set the initial Bias level to 0
+    void Initialize(void) override
+    {
+        dac_->SetPinFunction(kDacMode);
+        dac_->EnableDac(true);
+        dac_->SetAsAnalogMode(true);
+        dac_->SetPinMode(PinConfigureInterface::kInactive);
+        // Set Update Rate to 1MHz
+        dac_register->CR &= ~(1 << kBiasReg);
```
kammce: Since you have a bias function, you should use it here. You want to minimize code duplication and maximize code reuse, everywhere you can.

firmware/examples/Dac/dac.hpp
```diff
@@ -0,0 +1,99 @@
...
+    bool SetVoltage(float voltage) override
+    {
+        // value = (dac_out * 1024)/VrefP
+        float value = (voltage * 1024) / kVref;
+        int   conversion = static_cast<int>(value);
+        SJ2_ASSERT_FATAL(voltage < kVref,
+                         "Input voltage above SJ2"
+                         "output capacity of 3.3 Volts.");
+        dac_register->CR &= ~(kClearDac << kDacOutReg);
+        dac_register->CR |= (conversion << kDacOutReg);
```
kammce: Since you have a `WriteDac` function, you should use it here. You want to minimize code duplication and maximize code reuse, everywhere you can.

firmware/examples/Dac/dac.hpp
```diff
@@ -0,0 +1,99 @@
...
+    // , max current, and the allowed maximum update rate
+    void SetBias(BiasLevel bias_level) override
+    {
+        uint8_t bias = static_cast<int>(bias_level);
+        dac_register->CR &= ~(bias << kBiasReg);
+        dac_register->CR |= (bias << kBiasReg);
+    }
+ private:
+    PinConfigureInterface * dac_;
+    PinConfigure  dac_pin_;
```
kammce: Remove the extra space between PinConfigure.

firmware/examples/Dac/dac_test.cpp
```diff
@@ -0,0 +1,89 @@
...
+TEST_CASE("Testing Dac", "[dac]")
+{
+    using fakeit::Fake;
+    using fakeit::Mock;
+    using fakeit::Spy;
+    using fakeit::Verify;
+    using fakeit::When;
+    using fakeit::Using;
+
+    LPC_IOCON_TypeDef local_iocon;
```
kammce: Remove this since you are not using this.

firmware/examples/Dac/dac_test.cpp
```diff
@@ -0,0 +1,89 @@
...
+    memset(&local_iocon, 0, sizeof(local_iocon));
```
kammce: Remove this since you are not using this.

firmware/examples/Dac/dac_test.cpp
```diff
@@ -0,0 +1,89 @@
...
+    LPC_DAC_TypeDef local_dac_port;
+    Dac::dac_register = &local_dac_port;
+    PinConfigure iocon_pin_config(0, 26);
+    PinConfigure::pin_map =
+        reinterpret_cast<PinConfigure::PinMap_t *>(&local_iocon);
```
kammce: Remove this since you are not using this.

firmware/examples/Dac/project_config.hpp
```diff
@@ -0,0 +1,10 @@
+// This file overrides the default configuration options in the
```
kammce: I'm going to go through the examples and remove this project_config.hpp file. All of them do not depend on custom modification in this file, therefor the are redundant.

kammce: Please remove this from your example Dac.

firmware/examples/Dac/source/main.cpp
kammce: Correct.

firmware/examples/Dac/source/main.cpp
```diff
@@ -0,0 +1,156 @@
...
+     TriangleDac,
+     SawtoothDac,
+     SineDac
+};
+void StartDemo(Dac dac_demo, uint32_t input_cycles)
+{
+    for (int j = 0; j < 60; j++)
+    {
+        dac_controllers[0](&dac_demo, input_cycles);
+    }
```
kammce: Lol. So the reason why I said to use a look up table, is so you only have to create two for loops. You shouldn't be repreating them multiple times.

kammce:
```
for (int i = 0; i < sizeof(dac_controllers)/sizeof(*dac_controllers); i++)
{
    for (int j = 0; j < 60; j++)
    {
        dac_controllers[i](&dac_demo, input_cycles);
    }
}
```

kammce: Can you also consider reducing the time in each waveform? Because it should be a quick demo someone can run through within a minute of having everything connected. 60 iterations is super long, especially when you are waiting to see the last waveform.

firmware/examples/Dac/source/main.cpp
```diff
@@ -0,0 +1,148 @@
...
+    serrated = 0;
+    for ( uint8_t i = 127; i < 253 ; i++ )
+    {
+        uint16_t conversion = static_cast<uint16_t>(serrated);
+        dac_demo->WriteDac(conversion);
+        serrated = serrated + 4;
+        Delay(delay);
+    }
+}
+void (*dac_controllers[4])(Dac *, uint32_t) = {
```
kammce: You should be able to omit the number 4 here. Since the compiler should be able to deduce the size of the lookup table.

firmware/examples/Dac/source/main.cpp
```diff
@@ -0,0 +1,148 @@
...
+     SerratedDac,
+     TriangleDac,
+     SawtoothDac,
+     SineDac
+};
```
kammce: Add a newline below this point.

RadConMannBor: By newline do you just mean blank? or like  ' /n '

firmware/library/L1_Drivers/dac.hpp
```diff
@@ -1,19 +1,98 @@
...
+        dac_register->CR &= ~(kClearDac << kDacOutReg);
+        dac_register->CR |= (conversion << kDacOutReg);
+        return true;
+    }
+    // Sets the Bias for the Dac, which determines the settling time
+    // , max current, and the allowed maximum update rate
+    void SetBias(BiasLevel bias_level) override
+    {
+        uint8_t bias = static_cast<int>(bias_level);
+        dac_register->CR |= (bias << kBiasReg);
```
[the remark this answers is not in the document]
kammce: This problem has not been solved. You need to clear the bit then set it with the OR operator.

firmware/library/L1_Drivers/dac_test.cpp
```diff
@@ -0,0 +1,89 @@
...
+    SECTION("SetBias")
+    {
+        // Source: "UM10562 LPC408x/407x User manual" table 686 page 814
+        constexpr uint8_t kBiasMask = 0b0;
+        constexpr uint8_t kBias0    = 0b0;
+        constexpr uint8_t kBias1    = 0b1;
+        local_dac_port.CR           = kBiasMask;
+        // Test Setting Bias to 0 and checking if the register is correct
+        test_subject00.SetBias(Dac::BiasLevel::kBiasHigh);
+        CHECK((local_dac_port.CR >> Dac::kBiasReg) == kBias0);
```
kammce: This shouldn't be a shift. It should be an AND function. This goes for all tests. See the other tests by side effect examples like pin configure and pwm.

RadConMannBor: Oh your right, I changed it to just be a delay so it shouldn't be long

kammce:
Resolves issue #4

PR #159, opened 2018-08-10
Changing from using python2 to python2.7

kammce:
Using python2.7 rather than python2 in 'lint' make target.
Python2.7 should be the true path that python2 is aliased with. For
systems that do not have python2.7 aliased with python2, this fix should
fix their issues.


kammce:
Covered by #172.

PR #164, opened 2018-08-14
SSP Implementation

diarmadhi:
SSP API interface with implementation

### DIFF
--- a/firmware/library/L1_Drivers/ssp.hpp
+++ b/firmware/library/L1_Drivers/ssp.hpp
@@ -8,7 +8,8 @@
 //          SSP1    |   P0.8    |   P0.9    |   P0.7
 #pragma once
 
-#include <cstdint>
+#include "L0_LowLevel/LPC40xx.h"
+#include "L1_Drivers/pin_configure.hpp"
 
 class SspInterface
 {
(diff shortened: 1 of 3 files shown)


diarmadhi:
Understood. I'm still having errors with my tests and figured I should get the implementation to start reviewing while I figure out what I'm doing wrong with the tests.

diarmadhi:
Note that the error with Travis-ci is that clang-tidy is "not installed" for the first test. On the second test, tidy is throwing the same "non-const" error for the unit test constructor. 

Otherwise, everything else passes. This includes working unit tests and an example that works with external SPI flash memory.

firmware/library/L1_Drivers/ssp.hpp
```diff
@@ -70,3 +72,215 @@ class SspInterface
...
+    static constexpr uint8_t kDataBit = 0;
+    static constexpr uint8_t kFrameBit = 4;
+    static constexpr uint8_t kPolarityBit = 6;
+    static constexpr uint8_t kPhaseBit = 7;
+    static constexpr uint8_t kDividerBit = 8;
+    static constexpr uint8_t kMasterModeBit = 2;
+    static constexpr uint8_t kDataLineIdleBit = 4;
+
+    // Default constructor sets up SSP0 peripheral as SPI master
+    constexpr Ssp() : mosi_(mosi_pin_),
```
kammce: Put mosi on a newline.

firmware/library/L1_Drivers/ssp.hpp
```diff
@@ -70,3 +72,215 @@ class SspInterface
...
+        miso_(miso_pin_),
+        sck_(sck_pin_),
+
+        mosi_pin_(kPort0, kPin18),   // P0.18
+        miso_pin_(kPort0, kPin17),   // P0.17
+        sck_pin_(kPort0, kPin15),    // P0.15
```
kammce: I think that the comments can be omitted since its kinda of obvious from the variable names.

firmware/library/L1_Drivers/ssp.hpp
```diff
@@ -70,3 +72,215 @@ class SspInterface
...
+
+        pssp_(Peripheral::kSsp0), pconp_bits_(kPconp0), master_mode_(kMaster),
```
kammce: Please use a newline for each of the initializer list items, unless clang-format did this.

firmware/library/L1_Drivers/ssp.hpp
```diff
@@ -70,3 +72,215 @@ class SspInterface
...
+            ssp_registers[static_cast<uint32_t>(pssp_)]->CPSR |=
+                (clock_prescaler_);
+
+            // Enable SSP pins
+            mosi_.SetPinFunction(kEnableSspPins);
+            miso_.SetPinFunction(kEnableSspPins);
+            sck_.SetPinFunction(kEnableSspPins);
+
+            // Clear and Set Control Register values
+            ssp_registers[static_cast<uint32_t>(pssp_)]->CR0 &=
```
kammce: Rather than having to write `static_cast` multiple times, create a local variable.

firmware/library/L1_Drivers/ssp.hpp
```diff
@@ -70,3 +72,215 @@ class SspInterface
...
+    PowerOn pconp_bits_;
+    MasterSlaveMode master_mode_;
+    DataSize data_size_;             // data size, number of bits
+    FrameMode frame_format_;         // frame format
+    bool clock_polarity_;            // clock polarity
+    bool clock_phase_;               // clock phase
+    // clock_divider_ is the number of prescaler-output clocks per bit
+    // see page 611 of UM10562 for SCR-Serial Clock Rate for more info
+    uint8_t clock_divider_;
+    uint8_t clock_prescaler_;        // clock prescaler
```
kammce: Ok, you may hate me for doing this, but rather than holding these variables within the class, you can save memory by simply use the registers. 

You should create getters and setters for each of these modes, by reading from the registers. This is also helpful, because the registers are the ground truth for the values. There is potential for your class to be out of sync with the registers.

The conclusion, is don't store these values in class variables, if you can derive their values from the registers.

firmware/library/L1_Drivers/ssp_test.cpp
```diff
@@ -0,0 +1,81 @@
...
+
+    SECTION("Test Pin Numbers")
+    {
+        fakeit::Verify(Method(
+            mock_mosi, SetPinFunction).Using(Ssp::kEnableSspPins));
+        fakeit::Verify(Method(
+            mock_miso, SetPinFunction).Using(Ssp::kEnableSspPins));
+        fakeit::Verify(Method(
+            mock_sck, SetPinFunction).Using(Ssp::kEnableSspPins));
+    }
```
kammce: Your tests need to include a test for every single method in your class, to get full coverage.

firmware/library/L1_Drivers/ssp.hpp
diarmadhi: This approach would essentially break apart the Initialize() function wouldn't it? Since it is no longer using these variables to initialize everything at once? 

I think I will also have to redo each constructor as well? For example, the default constructor sets everything up to be SPI master when Initialize() runs. If I remove all these variables, it won't have the default ability anymore...There are a few things I can automatically set given the peripheral option (SSP0 vs SSP1) and just make Initialize() the power up function... Hmm, let me see what I can do.

firmware/library/L1_Drivers/ssp.hpp
```diff
@@ -61,12 +63,227 @@ class SspInterface
...
+    static constexpr uint8_t kPin9 = 9;
+    static constexpr uint8_t kPin8 = 8;
+    static constexpr uint8_t kPin7 = 7;
+    static constexpr uint8_t kDataBit = 0;
+    static constexpr uint8_t kFrameBit = 4;
+    static constexpr uint8_t kPolarityBit = 6;
+    static constexpr uint8_t kPhaseBit = 7;
+    static constexpr uint8_t kDividerBit = 8;
+    static constexpr uint8_t kMasterModeBit = 2;
+    static constexpr uint8_t kDataLineIdleBit = 4;
```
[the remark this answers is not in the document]
kammce: Nope, because they are super specific to this exact implementation. The interface should be usable across MCUs, CPUs ect. We should be able to swap out your SPI implementation with one meant for the PIC and it should still be able to properly reference the SpiInterface.

firmware/library/L1_Drivers/ssp.hpp
kammce: You correct. The idea here, would be to have Initialize simply enable the SPI and power it on. Or maybe having SPI initialize sets up some defaults and then you add an enable method to power on the SPI. If initialize was used just to power everything on, the code one would write would be the same. You would set all of the configurations using the appropriate methods, and call initialize at the end to enable the spi device. You could also introduce a `SetupDefaults` method that can be used on its own, or maybe even in the init function if none of the other methods have been called. Tell me what you think and what you think would be a cleaner design.

One thing, I don't think we should do is use internal variables for the class.

@SameerAzer What do you think?

diarmadhi: Would it be acceptable to group like items together, similar to my current setter functions, or would it be better to make an individual one for each?

firmware/library/L1_Drivers/ssp.hpp
```diff
@@ -70,3 +71,235 @@ class SspInterface
...
+    PinConfigureInterface & sck_;
+
+    PinConfigure mosi_pin_;
+    PinConfigure miso_pin_;
+    PinConfigure sck_pin_;
+
+    // SSP member variables
+    Peripheral pssp_;                // SSP interfaces
+
+// TODO(Khrysta): Replace following variables with get/set functions
```
kammce: Indent this, also, create a new issue and put the issue number in the (). For example TODO(#1234) where 1234 is the issue number. This is so, anyone else can pick up where you left off rather than going and finding you to fix this problem.

PR #169, opened 2018-08-15
Implementation file for Uart and example code.

Jeremy-Chau:
We left out the time limit parameter for now. If we can get it to properly work before the expo we will add it to our code.


kammce:
Closing stale PR.

PR #173, opened 2018-08-17
Showcase for SSP Peripheral

diarmadhi:
Example showing how SSP works as a SPI Master with
SPI flash memory.

### DIFF
--- a/firmware/examples/Ssp/source/main.cpp
+++ b/firmware/examples/Ssp/source/main.cpp
@@ -1,0 +1,65 @@
+#include <cstdarg>
+#include <cstdint>
+#include <cstdio>
+#include <cstdlib>
+
+#include "config.hpp"
+#include "L0_LowLevel/delay.hpp"
+#include "L0_LowLevel/LPC40xx.h"
+#include "L1_Drivers/gpio.hpp"
+// #include "L1_Drivers/pin_configure.hpp"
+#include "L1_Drivers/ssp.hpp"
+#include "L2_Utilities/debug_print.hpp"
+
+int main(void)
+{
+    uint16_t data;
+    // static constexpr read_id = 0x9F;
+    // This application sets up the SSP peripheral as a SPI master and will
+    // interact with the external SPI flash memory chip (AT25DN. Once the
+    // connection is established, the LPC master will request the device ID
+    // from the flash memory.
+    Delay(5000);
+    DEBUG_PRINT("SSP Application Starting...");
+
+    // Select SSP1 peripheral
+    Ssp ssp1_spi(SspInterface::Peripheral::kSsp1);
+    DEBUG_PRINT("Setting up peripheral pins: MOSI on Port 0.9.");
+    DEBUG_PRINT("MISO on Port 0.8, and SCK on Port 0.7.");
+    // Enable SSP1 peripheral power
+    ssp1_spi.SetPeripheralPower();
+    DEBUG_PRINT("Set power for SSP1 peripheral");
+    // Set SSP1 as SPI master
+    ssp1_spi.SetPeripheralMode(SspInterface::MasterSlaveMode::kMaster,
+                               SspInterface::FrameMode::kSpi,
+                               SspInterface::DataSize::kEight);
+    DEBUG_PRINT("Set SSP1 as SPI master.");
+    // Set up SPI clock polarity and phase
+    ssp1_spi.SetClock(1, 1, 2, 0);
+    DEBUG_PRINT("SCK will read high when inactive.");
+    DEBUG_PRINT("MOSI will read low when inactive.");
+    // Initialize SSP
+    DEBUG_PRINT("SSP initialization status: %d", ssp1_spi.Initialize());
+
+    // Set up chip select as GPIO pin
+    Gpio chip_select(0, 6);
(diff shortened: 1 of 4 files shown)


firmware/examples/Ssp/source/main.cpp
```diff
@@ -0,0 +1,65 @@
...
+    // Get Device ID from flash memory, read id command = 0x9f
+    DEBUG_PRINT("Observe the logic analyzer for SPI interaction.");
+    DEBUG_PRINT("Sending command D7h to read status register.");
+    chip_select.SetLow();
+    ssp1_spi.Transfer(0xD7);
+    data = ssp1_spi.Transfer(0);
+    chip_select.SetHigh();
+    DEBUG_PRINT("Flash Status Reg: 0x%.2x", data);
+    DEBUG_PRINT("Ready/Busy Status: %d", (data >> 7));
+    DEBUG_PRINT("Device ID: %.1x", ((data >> 2) & 0b1111));
```
kammce: Main shouldn't return. Throw a loop in here, or put a while loop at the end to indicate that the application is halting action. See pin configure's example for an example

firmware/examples/Ssp/source/main.cpp
```diff
@@ -0,0 +1,65 @@
+#include <cstdarg>
```
kammce: I don't think this is used. Can it be removed?

firmware/examples/Ssp/source/main.cpp
```diff
@@ -0,0 +1,65 @@
...
+#include <cstdint>
+#include <cstdio>
```
kammce: I don't think this being used either.

firmware/examples/Ssp/source/main.cpp
```diff
@@ -0,0 +1,65 @@
...
+#include <cstdlib>
```
kammce: Nor this.

firmware/examples/Ssp/source/main.cpp
```diff
@@ -0,0 +1,65 @@
...
+
+#include "config.hpp"
+#include "L0_LowLevel/delay.hpp"
+#include "L0_LowLevel/LPC40xx.h"
```
kammce: Not this either.

firmware/examples/Ssp/source/main.cpp
```diff
@@ -0,0 +1,65 @@
...
+#include "L1_Drivers/gpio.hpp"
+// #include "L1_Drivers/pin_configure.hpp"
```
kammce: Remove if commented out and not used.

firmware/examples/Ssp/source/main.cpp
```diff
@@ -0,0 +1,65 @@
...
+    // static constexpr read_id = 0x9F;
+    // This application sets up the SSP peripheral as a SPI master and will
+    // interact with the external SPI flash memory chip (AT25DN. Once the
+    // connection is established, the LPC master will request the device ID
+    // from the flash memory.
+    Delay(5000);
+    DEBUG_PRINT("SSP Application Starting...");
+
+    // Select SSP1 peripheral
+    Ssp ssp1_spi(SspInterface::Peripheral::kSsp1);
```
kammce: I think a name of simply, `spi` or `spi1` would be more readable.

firmware/examples/Ssp/source/main.cpp
```diff
@@ -0,0 +1,65 @@
...
+    DEBUG_PRINT("SSP initialization status: %d", ssp1_spi.Initialize());
+
+    // Set up chip select as GPIO pin
+    Gpio chip_select(0, 6);
+    chip_select.SetAsOutput();
+    chip_select.SetHigh();
+    Delay(20);
+    chip_select.SetLow();
+    Delay(20);
+    chip_select.SetHigh();
```
kammce: Why is the set low and set high necessary?

firmware/examples/Ssp/source/main.cpp
```diff
@@ -0,0 +1,65 @@
...
+    Delay(20);
+    DEBUG_PRINT("Chip select on Port 0.6 and initialized high.");
+
+    // Get Device ID from flash memory, read id command = 0x9f
+    DEBUG_PRINT("Observe the logic analyzer for SPI interaction.");
+    DEBUG_PRINT("Sending command D7h to read status register.");
+    chip_select.SetLow();
+    ssp1_spi.Transfer(0xD7);
```
kammce: Make this into a constexpr variable and give it the name indicating what it is doing.

firmware/examples/Ssp/source/main.cpp
kammce: You should also include a `0x` for the device id to indicate that it is also in hex.

firmware/examples/Ssp/source/main.cpp
kammce: Ah ok. Interesting. Is this the same for the flash memory on board the Sjsu one?

kammce:
Closing due to staleness.

kammce:
Closing stale PR. The example PR should be used for the spi flash memory onboard the SJTwo Rev.1.D.

PR #174, opened 2018-08-20
Need Help Reviewing

Adam-D-Sanchez:


firmware/library/L1_Drivers/onboardled.hpp
```diff
@@ -18,3 +20,81 @@ class OnBoardLedInterface {
...
+    Led2.SetAsOutput();
+    Led3.SetAsOutput();
+    // This block turns off all the LEDs by default, to save power.
+    led_states = 0xFF;
+    LedSetAll(led_states);
+  }
+
+  void LEDOn(uint8_t led_number) override
+  {
+    led_states = (led_states & ~(1 << led_number));
```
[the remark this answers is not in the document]
kammce: What happens if the user supplies a value greater than 3?

firmware/library/L1_Drivers/onboardled.hpp
```diff
@@ -18,3 +20,81 @@ class OnBoardLedInterface {
...
+  void LedSetAll(uint8_t value)
+  {
+  led_states = value;
+  if (led_states&(1 << Led0Position))
+  {
+  Led0.SetHigh();
+  }
+  else{
+  Led0.SetLow();
+  }
```
kammce: The code of the LedSetAll function should follow the standard in terms of proper indenting.

firmware/library/L1_Drivers/onboardled.hpp
```diff
@@ -18,3 +20,81 @@ class OnBoardLedInterface {
...
+
+
+  (led_states&(1 << Led1Position))? Led1.SetHigh(); : Led1.SetLow();
+  (led_states&(1 << Led2Position))? Led2.SetHigh(); : Led2.SetLow();
+  (led_states&(1 << Led3Position))? Led3.SetHigh(); : Led3.SetLow();
```
kammce: Why are these ternary expression when the above is a full if-else block. Would prefer if-else over ternary function calls. But even better would be a solution that doesn't require conditionals.

firmware/library/L1_Drivers/onboardled.hpp
```diff
@@ -18,3 +20,81 @@ class OnBoardLedInterface {
...
+  }
+
+  uint8_t LedGetValues(void)
+  {
+  return led_states;
+  }
+
+ protected:
+  uint8_t led_states;
```
kammce: Rather than using up space, even though this space consumption is quite small, it would be better to lookup the value from hardware. This way, we get the ground truth of the pin's state rather than a reflection of it from a variable that could be out of sync with it, if the pin's value is changed by some other set of code.

firmware/library/L1_Drivers/onboardled.hpp
```diff
@@ -18,3 +20,81 @@ class OnBoardLedInterface {
...
            virtual uint8_t LedGetValues(void)                        = 0;
 };
+
+class OnBoardLed: public Gpio, public OnBoardLedInterface
+{
+ public:
+  static constexpr uint8_t Led0Position = 0;
+  static constexpr uint8_t Led1Position = 1;
+  static constexpr uint8_t Led2Position = 2;
+  static constexpr uint8_t Led3Position = 3;
```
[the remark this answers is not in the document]
Adam-D-Sanchez: No, they aren't anymore. I completely forgot to remove them. Thank you for pointing that out.

firmware/library/L1_Drivers/button_driver.hpp
```diff
@@ -1,10 +1,10 @@
...
-};
+#pragma once
+
+class ButtonInterface {
+ public:
+         virtual void Init()                = 0;
+         virtual bool Read()                = 0;
+         virtual bool Released()            = 0;
+         virtual bool Pressed()             = 0;
+         virtual void InvertButtonSignal(bool enable_invert_signal = true) = 0;
```
kammce: If you have some means of changing how Button signals are interpreted, it would necessary to also have a getter for this. Otherwise, the developer will be guessing at whether or not a button had been inverted or not.

firmware/library/L1_Drivers/onboardled.hpp
```diff
@@ -18,3 +20,63 @@ class OnBoardLedInterface {
...
+  // then outputs. Once they are configured as such, they can be maniputlated
+  // as expected.
+  void Initialize(void) override
+    {
+    // This block will make sure all the GPIOs that were initialized in the
+    // last block will be set as outputs.
+    Led0.SetAsOutput();
+    // Led1.SetAsOutput();
+    // Led2.SetAsOutput();
+    // Led3.SetAsOutput();
```
kammce: Why are these commented out?

firmware/library/L1_Drivers/onboardled.hpp
```diff
@@ -18,3 +20,63 @@ class OnBoardLedInterface {
            virtual void LedSetAll(uint8_t value)                     = 0;
            virtual uint8_t LedGetValues(void)                        = 0;
 };
+
+class OnBoardLed: public OnBoardLedInterface
+{
+ public:
+  // Initialization involes setting the appropriate pins to be GPIO, and
+  // then outputs. Once they are configured as such, they can be maniputlated
+  // as expected.
```
kammce: This comment is unnecessary since the code speaks for itself. The user can easily read what the driver is doing from the code below.

firmware/library/L1_Drivers/onboardled.hpp
```diff
@@ -18,3 +20,63 @@ class OnBoardLedInterface {
...
+    Led0.SetAsOutput();
+    // Led1.SetAsOutput();
+    // Led2.SetAsOutput();
+    // Led3.SetAsOutput();
+    // This block turns off all the LEDs by default, to save power.
+    led_states = 0xFF;
+    LedSetAll(led_states);
+  }
+
+  void LedOn(uint8_t led_number) override
```
kammce: You can remove the Prefix of `LED` from all of the methods. Think about it this way, the developer will write some code like the following:

    OnboardLeds leds;
    leds.LedOn(1);

Having to write Led twice is unnecessary. The user can easily understand what it is that is happening if you omit it.

    OnboardLeds leds;
    leds.On(1);

Prefer to keep the method names full, but concise.

Adam-D-Sanchez: I will change it as so. I figured a more full name would work out better.

firmware/library/L1_Drivers/button_driver.hpp
Adam-D-Sanchez: I don't know how this commented ended up here.

firmware/library/L1_Drivers/onboardled.hpp
Adam-D-Sanchez: Ah, I realized I am doing a pull request on the wrong branch. I was trying to cut down on the number of errors. I should be pull requesting a different branch besides my experimental branch.

firmware/library/L1_Drivers/onboardled.hpp
Adam-D-Sanchez: Got it.

firmware/library/L1_Drivers/onboardled.hpp
```diff
@@ -76,6 +101,8 @@ class OnBoardLed: public OnBoardLedInterface
   }
 
  protected:
-  uint8_t led_states;
-  Gpio LED0(main_port_number, 1);
+  Gpio LED0(main_port_number, led0_port_specifier);
+  Gpio LED1(main_port_number, led1_port_specifier);
+  Gpio LED2(main_port_number, led2_port_specifier);
+  Gpio LED3(main_port_number, led3_port_specifier);
```
kammce: This could be an array of Gpio objects. No need to use a constexpr variable for the port and pin, you can put them in directly. This is because, the numbers aren't magic if it is obvious what they are.

Adam-D-Sanchez:
I see that there are some variances in between master and my branch, for some reason are slipping through the difftool that I have. I will get that fix right now and make a new commit that should be cleaner.

firmware/library/L1_Drivers/button_driver.hpp
Adam-D-Sanchez: I'm not working on this driver, but this ended up in my differences for some reason, I'll see if I can get this cleaned up right now.

PR #176, opened 2018-08-26
Pwm Example

delwin-lei:
Application of Pwm driver to power LEDs.

### DIFF
--- a/firmware/examples/PWM/source/main.cpp
+++ b/firmware/examples/PWM/source/main.cpp
@@ -1,0 +1,43 @@
+#include "L0_LowLevel/delay.hpp"
+#include "L1_Drivers/gpio.hpp"
+#include "L1_Drivers/pwm.hpp"
+#include "L2_Utilities/debug_print.hpp"
+
+int main(void)
+{
+    DEBUG_PRINT("Pwm Application Starting...");
+
+    DEBUG_PRINT("Creating Gpio powered LED...P4.28");
+    DEBUG_PRINT("This Gpio powered LED represents the maximum brightness");
+    DEBUG_PRINT("and is used as reference against the Pwm powered LED");
+    DEBUG_PRINT("during its sweep.");
+    DEBUG_PRINT("Connect a 470 ohm resistor between the power line and");
+    DEBUG_PRINT("the anode of the LED. The cathode is then connected to");
+    DEBUG_PRINT("P4.28 and P2.0(Pwm1).");
+    Gpio p4_28(4, 28);
+    p4_28.SetAsOutput();
+    p4_28.SetLow();
+
+    DEBUG_PRINT("Creating Pwm powered LED...P2.0 (Pwm1)");
+    Pwm p2_0 = Pwm::CreatePwm<1>();
+    // Initialize Pwm at 1,000 hz
+    uint32_t frequency = 1000;
+    p2_0.Initialize(frequency);
+    float duty = 0;
+
+    while (1)
+    {
+        while (frequency <= 5)
+        {
+            frequency = frequency/2;
+            p2_0.SetFrequency(frequency);
+            for (int i = 0; i <= 255; i++)
+            {
+                duty = static_cast<float>(i)/255.0;
+                p2_0.SetDutyCycle(duty);
+                Delay(10);
+            }
+        }
+    }
+    return 0;
+}


firmware/examples/PWM/source/main.cpp
```diff
@@ -0,0 +1,37 @@
...
+#include "L2_Utilities/debug_print.hpp"
+
+int main(void)
+{
+    DEBUG_PRINT("Pwm Application Starting...\n");
+
+    // Standard gpio powered LED
+    DEBUG_PRINT("Creating Gpio powered LED...\n");
+    Gpio p4_28(4,28);
+    p4_28.SetAsInput();
```
kammce: If you are setting up a Gpio to power on an LED, doesn't it need to be an output?

firmware/examples/PWM/source/main.cpp
```diff
@@ -0,0 +1,37 @@
...
+#include "L1_Drivers/gpio.hpp"
+#include "L1_Drivers/pwm.hpp"
+#include "L2_Utilities/debug_print.hpp"
+
+int main(void)
+{
+    DEBUG_PRINT("Pwm Application Starting...\n");
+
+    // Standard gpio powered LED
+    DEBUG_PRINT("Creating Gpio powered LED...\n");
```
kammce: This should explain why you are creating GPIO powered LED. When I first read this, I had no idea why this was included in this code. Remember that the audience reading this are CmpE 30 students (new at programming) and people unfamiliar with the framework. You need to actually spell out what it is that you are doing at each step, so that the user fully understands every portion of the code. 

See PinConfigure example as a template.

firmware/examples/PWM/source/main.cpp
```diff
@@ -0,0 +1,37 @@
...
+#include "L0_LowLevel/delay.hpp"
+#include "L1_Drivers/gpio.hpp"
+#include "L1_Drivers/pwm.hpp"
+#include "L2_Utilities/debug_print.hpp"
+
+int main(void)
+{
+    DEBUG_PRINT("Pwm Application Starting...\n");
+
+    // Standard gpio powered LED
```
kammce: This comment is not necessary. It can be read from the debug message below.

firmware/examples/PWM/source/main.cpp
kammce: You should also guide the user, to how they should setup the pin with an led and resistor. You could tell them to use a 470 ohm resistor for both gpio and pwm led. Then tell them that the GPIO represents the maximum brightness that the led can be.

firmware/examples/PWM/source/main.cpp
```diff
@@ -0,0 +1,37 @@
...
+    while(1)
+    {
+        for(duty = 0; duty < 1; duty += 0.01)
+        {
+            p2_0.SetDutyCycle(duty);
+            Delay(100);
+        }
+    }
+
+    DEBUG_PRINT("Halting any action.\n");
```
kammce: This line of code is unreachable, thus it shouldn't be here. Since you aren't halting action, you are looping above, you should state, above the while(1) loop that you are sweeping the duty cycle from 0 to 100%.

firmware/examples/PWM/source/main.cpp
```diff
@@ -0,0 +1,37 @@
...
+#include <cstdio>
+#include <cstdint>
+#include "L0_LowLevel/delay.hpp"
+#include "L1_Drivers/gpio.hpp"
+#include "L1_Drivers/pwm.hpp"
+#include "L2_Utilities/debug_print.hpp"
+
+int main(void)
+{
+    DEBUG_PRINT("Pwm Application Starting...\n");
```
kammce: Debug print already adds a newline character to the end of your string. The \n isn't needed here.

kammce:
https://wiki.sei.cmu.edu/confluence/display/c/FLP30-C.+Do+not+use+floating-point+variables+as+loop+counters

firmware/examples/PWM/source/main.cpp
```diff
@@ -0,0 +1,36 @@
+#include "L0_LowLevel/delay.hpp"
+#include "L1_Drivers/gpio.hpp"
+#include "L1_Drivers/pwm.hpp"
+#include "L2_Utilities/debug_print.hpp"
+
+int main(void)
+{
+    DEBUG_PRINT("Pwm Application Starting...");
+
+    DEBUG_PRINT("Creating Gpio powered LED...");
```
kammce: Print out which pin is being powered, so the user running this example doesn't have to look at the code to connect the hardware together.

firmware/examples/PWM/source/main.cpp
```diff
@@ -0,0 +1,36 @@
...
+    DEBUG_PRINT("This Gpio powered LED represents the maximum brightness");
+    DEBUG_PRINT("and is used as reference against the Pwm powered LED");
+    DEBUG_PRINT("during its sweep.");
+    // Connect a 470 ohm resistor between the power line and the anode
+    // of the LED. The cathode is then connected to the gpio and pwm pin.
```
kammce: Rather than comment this out, you should print this to the user so they know what to do.

firmware/examples/PWM/source/main.cpp
```diff
@@ -0,0 +1,36 @@
...
+    float duty = 0;
+
+    while (1)
+    {
+        for (int i = 0; i <= 10; i++)
+        {
+            duty = static_cast<float>(i)/10.0;
+            p2_0.SetDutyCycle(duty);
+            Delay(100);
+        }
```
kammce: Can you add another for loop to sweep through the frequency as well?

firmware/examples/PWM/source/main.cpp
```diff
@@ -0,0 +1,36 @@
...
+
+    DEBUG_PRINT("Creating Pwm powered LED...");
+    Pwm p2_0 = Pwm::CreatePwm<1>();
+    // Initialize Pwm at 1,000 hz
+    p2_0.Initialize(1'000);
+    float duty = 0;
+
+    while (1)
+    {
+        for (int i = 0; i <= 10; i++)
```
kammce: Can you make this more granular? Like between 0 -> 255, and decrease the Delay down to maybe 10ms or lower? This will make the fading effect look more smooth.

firmware/examples/PWM/source/main.cpp
delwin-lei: As it sweeps through the duty cycle to display the flickering?

PR #179, opened 2018-09-04
Initial Implementation of Onboard Leds

Adam-D-Sanchez:
This pull request will be cleaner since I couldn't get the other differences to go away, no matter what method I tried. This will now only take in the implementation file, which I will update separately from my experimental branch.


firmware/library/L1_Drivers/onboardled.hpp
```diff
@@ -7,14 +7,96 @@
...
+  kOn = true,
+  kOff = false
+  };
+           virtual void Initialize()                                      = 0;
+           virtual void On(uint8_t led_number)                            = 0;
+           virtual void Off(uint8_t led_number)                           = 0;
+           virtual void Set(uint8_t led_number,
+              OnBoardLedInterface::LightState state)                      = 0;
+           virtual void Toggle(uint8_t led_number)                        = 0;
+           virtual void SetAll(uint8_t value)                             = 0;
```
kammce: This could take the LightState enumeration as an input.

firmware/library/L1_Drivers/onboardled.hpp
```diff
@@ -7,14 +7,96 @@
...
+  enum LightState : bool
+  {
+  kOn = true,
+  kOff = false
+  };
+           virtual void Initialize()                                      = 0;
+           virtual void On(uint8_t led_number)                            = 0;
+           virtual void Off(uint8_t led_number)                           = 0;
+           virtual void Set(uint8_t led_number,
+              OnBoardLedInterface::LightState state)                      = 0;
```
kammce: You don't have to prefix this with `OnBoardLedInterface::` since you are within scope of that namespace.

firmware/library/L1_Drivers/onboardled.hpp
```diff
@@ -7,14 +7,96 @@
...
 class OnBoardLedInterface {
  public:
-           virtual void Initialize()                                 = 0;
-           virtual void LedOn(uint8_t led_number)                    = 0;
-           virtual void LedOff(uint8_t led_number)                   = 0;
-           virtual void LedSet(uint8_t led_number, bool on)          = 0;
-           virtual void LedToggle(uint8_t led_number)                = 0;
-           virtual void LedSetAll(uint8_t value)                     = 0;
-           virtual uint8_t LedGetValues(void)                        = 0;
+  enum LightState : bool
```
kammce: Make this `enum class` and indent this plox.

firmware/library/L1_Drivers/onboardled.hpp
```diff
@@ -7,14 +7,96 @@
...
+      for (uint8_t i = 0; i < 4; i++)
+      {
+        led[i].SetAsOutput();
+        led[i].SetHigh();
+      }
+    }
+
+  void On(uint8_t led_number) override
+  {
+    if (led_number > 3) {printf("Input Led number can't be greater than 3.");}
```
kammce: Rather than running your own printf case evaluation use the SJ2_ASSERT_WARNING(led < 4, "Input Led number can't be greater than 3, input = %d.", led_number);

kammce: That macro can be found in the L2/Macros.hpp file. Also, do the same for the other methods.

firmware/library/L1_Drivers/onboardled.hpp
```diff
@@ -7,14 +7,96 @@
...
+
+  // This function takes in the 4 least significant bits from value, and sets
+  // the led to be ON or OFF. The least significant bit corresponds to LED0,
+  // next least significant corresponds to LED1, etc. ON =1, OFF =0. The four
+  // most significant bits will be unused.
+  void SetAll(uint8_t value) override
+  {
+    for (uint8_t i = 0; i < 4; i++)
+    {
+      ((value >> i)&1) ? led[i].SetLow() : led[i].SetHigh();
```
kammce: For better code reuse, you should call the `Set` method for each led.

firmware/library/L1_Drivers/onboardled.hpp
Adam-D-Sanchez: My original vision of Toggle would only "flip" the state of the LED based on the given number. My original vision of SetAll() would take in an actual value. For example 0b0000'1010, which would make LED3 and LED1 on, but LED2 and LED0 off. I'm not sure how I could make that enumeration work in either case. I am open to suggestions on how to make this more clear for someone who will use this library.

firmware/library/L1_Drivers/onboardled.hpp
Adam-D-Sanchez: This has been fixed.

firmware/library/L1_Drivers/onboardled.hpp
Adam-D-Sanchez: This has been fixed.

firmware/library/L1_Drivers/onboardled.hpp
Adam-D-Sanchez: Fixed

firmware/library/L1_Drivers/onboardled.hpp
```diff
@@ -7,14 +7,99 @@
...
+      "Input Led number can't be greater than 3, input = %d.\n", led_number);
+    (state == LightState::kOn) ? On(led_number) : Off(led_number);
+  }
+
+  void Toggle(uint8_t led_number) override
+  {
+    SJ2_ASSERT_WARNING(led_number > 3,
+      "Input Led number can't be greater than 3, input = %d.\n", led_number);
+    bool swap = led[led_number].ReadPin();
+    (swap) ? Set(led_number, LightState::kOn):Set(led_number, LightState::kOff);
```
kammce: Ternery statements are meant to return something. Prefer to use if and else statements rather than ternery statments, in cases where you are merely executing code. ternery's used in this way are less readable then if-else.

firmware/library/L1_Drivers/onboardled.hpp
```diff
@@ -7,14 +7,99 @@
...
+      for (uint8_t i = 0; i < 4; i++)
+      {
+        led[i].SetAsOutput();
+        led[i].SetHigh();
+      }
+    }
+
+  void On(uint8_t led_number) override
+  {
+    SJ2_ASSERT_WARNING(led_number > 3,
```
kammce: Sorry for the back and forth, but looking at this after a nights sleep, I realize that this will cause a hard-fault or out of array access if you merely `ASSERT_WARNING`. Assert warning allows the code to continue be emits a warning message. `SJ2_ASSERT_FATAL` would be better since it will halt the program at that point.

firmware/library/L1_Drivers/onboardled.hpp
```diff
@@ -7,14 +7,99 @@
...
-           virtual void LedOff(uint8_t led_number)                   = 0;
-           virtual void LedSet(uint8_t led_number, bool on)          = 0;
-           virtual void LedToggle(uint8_t led_number)                = 0;
-           virtual void LedSetAll(uint8_t value)                     = 0;
-           virtual uint8_t LedGetValues(void)                        = 0;
+  enum class LightState : bool
+  {
+      kOn = true,
+      kOff = false
+  };
```
kammce: Still need to indent this.

Adam-D-Sanchez:
I noticed there's an error on line 2414 on the OSX test in Travis. Is that supposed to be there?

I also don't think it builds the file that I added.

I was kinda curious to see how it went, so I looked at each line as it was being generated.

Adam-D-Sanchez:
Because the Travis build #936 failed as of a stall, I did the following to retrigger the build:

`git commit --amend --no-edit` then `git push -f`. Hopefully that didn't mess anything up.

Adam-D-Sanchez:
I spent a little bit of time reinstalling everything, but finally got clang-format to work. I noticed that it still kept the array at the end on one line. Let me know if there are still formatting issues.

PR #181, opened 2018-09-05
Utility Namespace Testing and Implementation

hangbogu:


firmware/library/L2_Utilities/utility.cpp
```diff
@@ -0,0 +1,94 @@
...
+}
+uint32_t utility::Map(uint32_t value, uint32_t current_min,
+                      uint32_t current_max, uint32_t new_min, uint32_t new_max)
+{
+    int64_t map_value       = value;
+    int64_t map_current_min = current_min;
+    int64_t map_current_max = current_max;
+    int64_t map_new_min     = new_min;
+    int64_t map_new_max     = new_max;
+    return static_cast<uint32_t>((map_value - map_current_min) *
```
kammce: There is an issue with using ints and implicitly casting them into floats and back to ints. There may be truncation through out the operations which can lead to improper values. 

The best course of action is to make an overloaded version of this function that takes floats and inputs and outputs a float. The version of map that uses ints would handle the conversion to floats and simply call the float version.

Within the floating point version you should convert everything to doubles (use doubles rather than floats to improve precision during mathematical operations) and cast down to float at the end (reduce size at the end).

firmware/library/L2_Utilities/utility_test.cpp
```diff
@@ -0,0 +1,104 @@
...
+    {
+        utility::SetRandomSeed();
+        uint32_t randomtester1 = std::rand() % 100;
+        utility::SetRandomSeed();
+        uint32_t randomtester2 = utility::Random(0, 99);
+        CHECK(randomtester1 == randomtester2);
+    }
+}
+
+TEST_CASE("Testing Bit Manipulations", "[Bit Manipulation methods]")
```
kammce: keep the [] test names short. For example this should just be `[bit manipulation]`

firmware/library/L2_Utilities/utility_test.cpp
```diff
@@ -0,0 +1,104 @@
...
+{
+    SECTION("BitExtract")
+    {
+        CHECK(utility::BitExtract(15, 2, 3) == 7);
+        CHECK(utility::BitExtract(15, 3, 3) == 7);
+        CHECK(utility::BitExtract(15, 31, 32) == 15);
+        CHECK(utility::BitExtract(15, 0, 1) == 1);
+        CHECK(utility::BitExtract(15, 31, 28) == 0);
```
kammce: Rather than use the decimal version of these numbers, it would be more readable to write the values in binary. For example, `0b1111`, to illustrate that the value contains only 1s.

kammce: It would also be great to include some tests with numbers that only include all 1s.

firmware/library/L2_Utilities/utility_test.cpp
```diff
@@ -0,0 +1,104 @@
...
+    }
+
+    SECTION("BitPlace")
+    {
+        CHECK(utility::BitPlace(15, 2, 7, 3) == 15);
+        CHECK(utility::BitPlace(15, 4, true, 1) == 31);
+        CHECK(utility::BitPlace(15, 3, false, 1) == 7);
+        CHECK(utility::BitPlace(0xF0F0F0F0, 31, 0x0F0F0F0F, 32) == 0xFFFFFFFF);
```
kammce: The behavior of bit place should be to replace the set of bits with a new set of bits for a specific width and position. For example, lets say I have the hex value 0x12345678, and I want to place DEAD in the middle (8 bits the left), the result should be 0x12DEAD78. It shouldn't simply or the bits together. Otherwise, a simple bit set function would be a better implementation for this.

firmware/library/L2_Utilities/utility.cpp
```diff
@@ -0,0 +1,94 @@
...
+    for (uint8_t i = 0; i < width; i++)
+    {
+        temp = utility::BitClear(target, uint8_t(start_bit - i));
+    }
+    uint32_t mask = new_bits;
+    mask <<= (start_bit - (width - 1));
+    return temp | mask;
+}
+
+uint32_t utility::BitSet(uint32_t target, uint8_t bit_position)
```
kammce: Would be useful to put a `bit_width` parameter for this that defaults to 1. So the user can set multiple bits.

firmware/library/L2_Utilities/utility.cpp
```diff
@@ -0,0 +1,94 @@
...
+{
+    uint32_t mask = 1 << bit_position;
+    return target | mask;
+}
+uint32_t utility::BitClear(uint32_t target, uint8_t bit_position)
```
kammce: Would be useful to put a `bit_width` parameter for this that defaults to 1. So the user can clear multiple bits.

firmware/library/L2_Utilities/utility_test.cpp
```diff
@@ -0,0 +1,104 @@
+#include "L2_Utilities/utility.hpp"
```
kammce: Through out this whole file: expected value should be on the left side of the check expression. 

    CHECK(expected == FunctionUnderTest(param1, paramN));

firmware/library/L2_Utilities/utility_test.cpp
```diff
@@ -0,0 +1,104 @@
...
+        CHECK(randomtester1 == randomtester2);
+    }
+
+    SECTION("Random")
+    {
+        utility::SetRandomSeed();
+        uint32_t randomtester1 = std::rand() % 100;
+        utility::SetRandomSeed();
+        uint32_t randomtester2 = utility::Random(0, 99);
+        CHECK(randomtester1 == randomtester2);
```
kammce: Should check that the next few random numbers do not equal the first number.

kammce: Should also check that random numbers fall within the range, Run a for loop for 1000 numbers within a really narrow size, lets say 4 bits, or between 1 and 15. Make sure that the number never above 15 or below 1.

firmware/library/L2_Utilities/utility_test.cpp
```diff
@@ -0,0 +1,104 @@
...
+        CHECK(utility::BitRead(15, 3) == 1);
+        CHECK(utility::BitRead(15, 4) == 0);
+        CHECK(utility::BitRead(1, 0) == 1);
+        CHECK(utility::BitRead(0, 0) == 0);
+        CHECK(utility::BitRead(0x80000000, 31) == 1);
+        CHECK(utility::BitRead(0, 31) == 0);
+    }
+}
+
+TEST_CASE()
```
kammce: Looks like you forgot to give this fella a name.

firmware/library/L2_Utilities/utility_test.cpp
```diff
@@ -0,0 +1,104 @@
...
+    }
+
+    SECTION("Max")
+    {
+        CHECK(utility::Max(0xFFFFFFFF, 0) == 0xFFFFFFFF);
+        CHECK(utility::Max(0, 0) == 0);
+    }
+}
+
+TEST_CASE("Testing Map and Constrain methods", "[]")
```
kammce: A reasonable name for the `[]` would be `[map-constrain]`

firmware/library/L2_Utilities/utility.cpp
```diff
@@ -0,0 +1,94 @@
...
+uint32_t utility::Max(uint32_t compare_1, uint32_t compare_2)
+{
+    if (compare_2 >= compare_1){
+        return compare_2;
+    }
+    else{
+        return compare_1;
+    }
+}
+uint32_t utility::Map(uint32_t value, uint32_t current_min,
```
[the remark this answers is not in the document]
hangbogu: Addressed most of the other requests, but just figured out what you meant in the comment on this and the following request.

firmware/library/L2_Utilities/utility.cpp
```diff
@@ -0,0 +1,94 @@
...
+    mask          = ~mask;
+    return target & mask;
+}
+uint32_t utility::BitRead(uint32_t target, uint8_t bit_position)
+{
+    uint32_t mask       = 1 << bit_position;
+    uint32_t masked_var = target & mask;
+    return masked_var >> bit_position;
+}
+uint32_t utility::Min(uint32_t compare_1, uint32_t compare_2)
```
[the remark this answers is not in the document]
kammce: Can you template this function so it can be used with most primitive types?

firmware/library/L2_Utilities/utility.cpp
kammce: Can you also template this function as well. With appropriate testes for int, float and double.

firmware/library/L2_Utilities/utility_test.cpp
```diff
@@ -0,0 +1,126 @@
+#include "L2_Utilities/utility.hpp"
+#include "L5_Testing/testing_frameworks.hpp"
+
+TEST_CASE("Testing Random Functions", "[random]")
+{
+    SECTION("SetRandomSeed Consistency")
+    {
+        utility::SetRandomSeed();
+        uint32_t randomtester1 = std::rand() % 100;
+        utility::SetRandomSeed();
```
kammce: Just realized that you should have another set of conditional tests for SetRandomSeed where you actually give it a seed value and make sure that works as expected.

firmware/library/L2_Utilities/utility.cpp
```diff
@@ -0,0 +1,76 @@
...
+{
+    return compare_1 <= compare_2 ? compare_1 : compare_2;
+}
+uint32_t utility::Max(uint32_t compare_1, uint32_t compare_2)
```
kammce: Template this as well.

firmware/library/L2_Utilities/utility.cpp
```diff
@@ -0,0 +1,76 @@
+#include <cstdlib>
+
+#include "utility.hpp"
+
+uint32_t utility::Random(uint32_t min, uint32_t max)
```
kammce: This should actually be `int32_t`. Users may want to accept negative numbers.

hangbogu: Will do.

firmware/library/L2_Utilities/utility_test.cpp
```diff
@@ -0,0 +1,158 @@
...
+        CHECK(true == norandomoob);
+    }
+
+    SECTION("Random")
+    {
+        utility::SetRandomSeed();
+        int32_t randomtester1 = std::rand() % 100;
+        utility::SetRandomSeed();
+        int32_t randomtester2 = utility::Random(0, 99);
+        CHECK(randomtester1 == randomtester2);
```
hangbogu: I was not sure how to test for float equality in the tests.

firmware/library/L2_Utilities/utility_test.cpp
```diff
@@ -0,0 +1,158 @@
...
+{
+    SECTION("Map")
+    {
+        CHECK(1 == utility::Map(1, 1, 4, 1, 256));
+        CHECK(86 == utility::Map(2, 1, 4, 1, 256));
+        CHECK(1 == utility::Map(85, 1, 256, 1, 4));
+        CHECK(2 == utility::Map(86, 1, 256, 1, 4));
+        CHECK(4 == utility::Map(1, 1, 4, 4, 1));
+        CHECK(1 == utility::Map(4, 1, 4, 4, 1));
+
```
hangbogu: I wasn't sure how to test float equality.

kammce: You can test it by doing the following

    if(between - .01 <= value && value <= between + 0.01)

It would be nice to add another utility function that handles this for you.

kammce:
This PR is stale and the contributor hasn't been responding. Closing this PR.

PR #187, opened 2018-09-23
Updated HelloWorld Sample Project

kammce:
- Added a constexpr constructor to OnBoardLed() with initializer list
- Fixed warning in OnBoardLed::GetStates()
- HelloWorld no merely counts in binary on the LEDs.

### DIFF
--- a/firmware/HelloWorld/source/main.cpp
+++ b/firmware/HelloWorld/source/main.cpp
@@ -1,38 +1,26 @@
 #include <project_config.hpp>
 
 #include <cstdint>
-#include <cstdio>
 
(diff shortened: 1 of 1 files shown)


Adam-D-Sanchez:
I was looking at this code to see how you fixed that warning in onboard_led.hpp. I was going to ask you how you got that warning to stop appearing. I think the way you did it was pretty great and I hope it doesn't give us any further problems. However, I saw that the delay is outside the for loop though in your main.cpp. I think you meant to put it inside that for loop.

kammce:
@Adam-D-Sanchez Can you make a PR with just the warning fix added? That would be great!

PR #190, opened 2018-09-25
Adding TaskScheduler

HuangDave:
The TaskScheduler can be used to schedule and manage any tasks that inherits TaskInterface.
An example project can be found in the firmware/examples/TaskScheduler directory.

Other changes:
* Added default scheduler size constant in config.hpp
* Added vApplicationGetIdleTaskMemory to startup.hpp

### DIFF
--- a/firmware/examples/TaskScheduler/source/main.cpp
+++ b/firmware/examples/TaskScheduler/source/main.cpp
@@ -1,0 +1,62 @@
+#include <cstdarg>
+
+#include "L0_LowLevel/uart0.hpp"
+#include "L2_Utilities/debug_print.hpp"
+#include "L4_Application/task.hpp"
+
+class PrinterTask : public rtos::Task<1024>
+{
+ public:
+    PrinterTask(const char * task_name, const char * message)
+        : Task(task_name, rtos::Priority::kMedium),
+          message_(message),
+          run_count_(0){};
+    bool Setup() override
+    {
+        DEBUG_PRINT("Setup() for: %s", kName);
+        return true;
+    };
+    bool PreRun() override
+    {
+        DEBUG_PRINT("PreRun() for: %s", kName);
+        return true;
+    };
+    bool Run() override
+    {
+        run_count_ += 1;
+        DEBUG_PRINT("%s: %ld", message_, run_count_);
+        if (run_count_ == 10)
+        {
+            Delete();
+        }
+        return true;
+    };
+
+ private:
+    const char * message_;
+    uint32_t run_count_;
+};
+
+int main()
+{
+    DEBUG_PRINT("Starting TaskScheduler example...");
+    // TODO(#210): Should remove the use of new and use static allocation for
+    //             constructing tasks when the issue is fixed.
+    PrinterTask * printer_one =
+        new PrinterTask("Printer A", "I am a printer, I am faster");
+    printer_one->SetDelayTime(500);
+    PrinterTask * printer_two =
+        new PrinterTask("Printer B", "I am also a printer");
+    printer_two->SetDelayTime(1000);
+
+    DEBUG_PRINT("Attempting to search for Printer A in the scheduler...");
+    rtos::TaskInterface & task =
+        *(rtos::TaskScheduler::Instance().GetTask("Printer A"));
+    DEBUG_PRINT("Found task: %s", task.GetName());
+
+    rtos::TaskScheduler::Instance().Start();
+    DEBUG_PRINT("This point should not be reached!");
+    delete printer_one;
+    delete printer_two;
+    return 0;
+}
--- a/firmware/library/L0_LowLevel/startup.cpp
+++ b/firmware/library/L0_LowLevel/startup.cpp
@@ -243,10 +243,25 @@ IsrPointer dynamic_isr_vector_table[] = {
     EepromIrqHandler,       // 56, 0xe0 - EEPROM
 };
 
-
 extern "C" void vPortSetupTimerInterrupt(void)  // NOLINT
 {
     // Empty implementation, startup handles this itself.
+}
+
+static StaticTask_t idle_task_tcb;
+static StackType_t idle_task_stack[configMINIMAL_STACK_SIZE];
+// Implementation of vApplicationGetIdleTaskMemory required when
+// configSUPPORT_STATIC_ALLOCATION == 1.
+// The function is called to statically create the idle task when
+// vTaskStartScheduler is invoked.
+extern "C" void vApplicationGetIdleTaskMemory(  // NOLINT
+    StaticTask_t ** ppx_idle_task_tcb_buffer,
+    StackType_t ** ppx_idle_task_stack_buffer,
+    uint32_t * pul_idle_task_stack_size)
+{
+    *ppx_idle_task_tcb_buffer   = &idle_task_tcb;
+    *ppx_idle_task_stack_buffer = idle_task_stack;
+    *pul_idle_task_stack_size   = SJ2_ARRAY_LENGTH(idle_task_stack);
 }
 
 // .data Section Table Information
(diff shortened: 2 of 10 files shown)


kammce:
@HuangDave I forgot about the xTaskCreateStatic (https://www.freertos.org/xTaskCreateStatic.html) which can do task creation statically and during compile time. I would love to be able to handle this one primarily for the task scheduler. Can you also look into switching to this for this implementation?

firmware/library/L4_Application/task_scheduler_test.cpp
```diff
@@ -0,0 +1,129 @@
...
+FAKE_VOID_FUNC(vTaskDelete, TaskHandle_t);
+FAKE_VOID_FUNC(vTaskDelayUntil, TickType_t *, TickType_t);
+FAKE_VOID_FUNC(vApplicationGetIdleTaskMemory, StaticTask_t **, StackType_t **,
+               uint32_t *);
+FAKE_VALUE_FUNC(TickType_t, xTaskGetTickCount);
+FAKE_VALUE_FUNC(TaskHandle_t, xTaskCreateStatic, TaskFunction_t, const char *,
+                uint32_t, void *, UBaseType_t, StackType_t *, StaticTask_t *);
+FAKE_VALUE_FUNC(EventGroupHandle_t, xEventGroupCreateStatic,
+                StaticEventGroup_t *);
+FAKE_VALUE_FUNC(EventBits_t, xEventGroupSync, EventGroupHandle_t, EventBits_t,
```
HuangDave: Should these be place in a header file for reuse?

firmware/examples/TaskScheduler/source/main.cpp
```diff
@@ -0,0 +1,61 @@
...
+    {
+        run_count_ += 1;
+        DEBUG_PRINT("%s: %ld", message_, run_count_);
+        if (run_count_ == 10)
+        {
+            Delete();
+        }
+        return true;
+    };
+    PrinterTask(const char * task_name, const char * message)
```
kammce: This class should follow the c++ class layout: https://google.github.io/styleguide/cppguide.html#Class_Format
Class constructor should be on top.

firmware/examples/TaskScheduler/source/main.cpp
```diff
@@ -0,0 +1,61 @@
...
+
+ private:
+    const char * message_;
+    uint32_t run_count_;
+};
+
+int main()
+{
+    DEBUG_PRINT("Starting TaskScheduler example...");
+    PrinterTask * printer_one =
```
kammce: Could you have an example of you doing this without using new? 
You could put that version of the variable in global space so it has dedicated memory for it in the .data section.

firmware/examples/TaskScheduler/source/main.cpp
```diff
@@ -0,0 +1,61 @@
...
+    PrinterTask * printer_two =
+        new PrinterTask("Printer B", "I am also a printer");
+    printer_two->SetDelayTime(1000);
+
+    DEBUG_PRINT("Attempting to search for Printer A in the scheduler...");
+    rtos::TaskInterface & task =
+        *(rtos::TaskScheduler::Instance().GetTask("Printer A"));
+    DEBUG_PRINT("Found task: %s", task.GetName());
+
+    rtos::TaskScheduler::Instance().Start();
```
kammce: Does this quit when all tasks are ended?

firmware/library/L0_LowLevel/startup.cpp
```diff
@@ -243,12 +243,27 @@ IsrPointer dynamic_isr_vector_table[] = {
...
+// The function is called to statically create the idle task when
+// vTaskStartScheduler is invoked.
+extern "C" void vApplicationGetIdleTaskMemory(  // NOLINT
+    StaticTask_t ** ppx_idle_task_tcb_buffer,
+    StackType_t ** ppx_idle_task_stack_buffer,
+    uint32_t * pul_idle_task_stack_size)
+{
+    *ppx_idle_task_tcb_buffer   = &idle_task_tcb;
+    *ppx_idle_task_stack_buffer = idle_task_stack;
+    *pul_idle_task_stack_size   = configMINIMAL_STACK_SIZE;
```
kammce: This will become a problem if the top array's stack size is changed without changing this assignment here. A better alternative is to use `SJ2_ARRAY_LENGTH(idle_task_stack)`. This will always follow the `idle_task_stack` regardless of what size it is, thus you never have to change multiple values for everything to work.

firmware/library/L4_Application/task.hpp
```diff
@@ -0,0 +1,150 @@
+// This file contains the TaskInterface class.
+// All tasks must inherit TaskInterface and override the Run() function.
+// Usage:
+//      class PrinterTask : public Task<1024>;
+//      PrinterTask * printer_one =
```
kammce: Definitions here should also not use dynamic memory.

firmware/library/L4_Application/task_scheduler_test.cpp
kammce: Yes, they should be. The rtos.hpp file would be a great place to put these. 

You can do something like this:
```
#if HOST_TEST
// insert the task fakes
#else
#include "FreeRTOS.h"
#endif  // HOST_TEST 
```

firmware/library/L4_Application/task_scheduler.cpp
```diff
@@ -0,0 +1,109 @@
...
+    const EventBits_t kPreRunSyncBits =
+        TaskScheduler::Instance().GetPreRunSyncBits();
+    const uint32_t kSyncBit = (1 << kTaskIndex);
+    SJ2_ASSERT_FATAL(task.PreRun(),
+                     "PreRun() failed for task: %s, terminating scheduler!",
+                     task.GetName());
+    // wait for all other PreRun() of other tasks to finish
+    xEventGroupSync(pre_run_event_group_handle, kSyncBit, kPreRunSyncBits,
+                    portMAX_DELAY);
+    // ALl PreRun() complete, each Task's Run() can now start executing...
```
kammce: Nit: misspelled ALl -> All

firmware/library/L4_Application/task_scheduler.cpp
```diff
@@ -0,0 +1,109 @@
...
+    TickType_t last_wake_time = xTaskGetTickCount();
+    while (1)
```
kammce: Nit: prefer `true` vs `1`.

firmware/library/L4_Application/task_scheduler.cpp
```diff
@@ -0,0 +1,109 @@
...
+    vTaskStartScheduler();
+};
+
+void rtos::TaskScheduler::InitializeAllTasks()
+{
+    for (uint32_t i = 0; i < task_count_; i++)
+    {
+        TaskInterface * task = task_list_[i];
+        *(task->GetHandle()) = xTaskCreateStatic(
+            RunTask, task->GetName(),
```
kammce: Nit: I think that having each parameter in its own line would look cleaner, even if `clang-format` disagrees.

firmware/library/L4_Application/task_scheduler.cpp
```diff
@@ -0,0 +1,109 @@
...
+    }
+    vTaskDelete(task_list_[kTaskIndex]->GetHandle());
+    // remove the task and update the task_list_ and task_count_
+    for (uint8_t i = kTaskIndex; i < task_count_; i++)
+    {
+        task_list_[i] = task_list_[i + 1];
+    }
+    task_count_--;
+    task_list_[task_count_] = nullptr;
+};
```
kammce: So, this implementation is fine. But I don't think we need to re-order the list each time something is removed. We could simply fill up task array items that are set to `nullptr`. If all tasks spaces are defined, then we know the task list is full.

firmware/examples/TaskScheduler/source/main.cpp
HuangDave: Without the new, a hard fault occurs.

kammce: Ok. Will look past this for this commit. Can you add an issue in the issues section, as well as a TODO in your code stating that we should move away from the usage of new here when we find a fix.

PR #203, opened 2018-10-02
Button Implementation and Unit Test update

Lawrence-Wan:
- Implementation of onboard button driver
- Added unit testing file for onboard button driver
- Removed all coverage files from commit


firmware/library/L1_Drivers/button_driver.hpp
```diff
@@ -1,10 +1,87 @@
...
+{
+ public:
+         virtual void Initialize()                        = 0;
+         virtual bool Read()                              = 0;
+         virtual bool Released()                          = 0;
+         virtual bool Pressed()                           = 0;
+         virtual void InvertButtonSignal(bool enable_invert_signal = true)  = 0;
+};
+
+class ButtonDriver : public ButtonInterface
```
kammce: Would prefer the name `Button` rather than `ButtonDriver`. No other classes are named `Driver`.

firmware/library/L1_Drivers/button_driver.hpp
```diff
@@ -1,10 +1,87 @@
...
+           : port_(kPortFive),
+             pin_(kPinFour),
+             gpio_(gpio),
+             pincon_(pincon),
+             gpio_pin_(port_, pin_),
+             pincon_pin_(Pin::CreateInactivePin())
+       {
+       }
+
+void Initialize() override {
```
kammce: Curly braces should be on newline.

firmware/library/L1_Drivers/button_driver.hpp
```diff
@@ -1,10 +1,87 @@
...
+bool Released() override {
+    return (Read()) ? true : false;
+}
+
+bool Pressed() override {
+    return (Read()) ? false : true;
+}
+
+void InvertButtonSignal(bool enable_invert_signal = true) override {
+  if (enable_invert_signal == false) {
```
kammce: Comparison not needed. You can simply put the variable into the `SetAsActiveLow()` function.

firmware/library/L1_Drivers/button_driver.hpp
```diff
@@ -1,10 +1,87 @@
...
+             gpio_(gpio_pin_),
+             pincon_(pincon_pin_),
+             gpio_pin_(port_, pin_),
+             pincon_pin_(port_, pin_)
+       {
+       }
+
+       // Unit test constructor
+       constexpr ButtonDriver(const GpioInterface & gpio,
+                              const PinInterface & pincon)
```
kammce: Gpio extends Pin. No need to take in pincon.

firmware/library/L1_Drivers/button_driver.hpp
```diff
@@ -1,10 +1,87 @@
 #pragma once
```
kammce: This file should be in L3_HAL.

firmware/library/L1_Drivers/button_driver.hpp
```diff
@@ -1,10 +1,87 @@
...
+      pincon_pin_.SetMode(PinInterface::Mode::kInactive);
+      pincon_pin_.SetPinFunction(kFunction0);
+      gpio_pin_.SetAsInput();
+}
+
+bool Read() override {
+    return (gpio_pin_.ReadPin()) ? true : false;
+}
+
+bool Released() override {
```
kammce:
```
private: 
  static bool was_pressed = false;
public:
bool Released(bool reset_state = false)
{
    bool result = false;
    if(reset_state)
    {
        was_pressed = false;
    }
    if(ReadPin() && !was_pressed)   
    {
        was_pressed = true;
    }
    else if(!ReadPin() && was_pressed)
    {
        result = true;
        was_pressed = false;
    }
    return result;
}


Button btn(2, 5);

while(!btn.Released())
{
    continue;
}

btn.Released(false);
```

This is what this thing should do.

firmware/library/L1_Drivers/button_driver.hpp
```diff
@@ -1,10 +1,87 @@
...
+    return (Read()) ? true : false;
+}
+
+bool Pressed() override {
```
kammce: Similar for the Pressed case.

firmware/library/L1_Drivers/button_driver.hpp
```diff
@@ -1,10 +1,87 @@
...
+             pin_(kPinFour),
+             gpio_(gpio),
+             pincon_(pincon),
+             gpio_pin_(port_, pin_),
+             pincon_pin_(Pin::CreateInactivePin())
+       {
+       }
+
+void Initialize() override {
+      pincon_pin_.SetMode(PinInterface::Mode::kInactive);
```
kammce: Rev1.D requires internal pulldown to work.

firmware/library/L1_Drivers/button_driver.hpp
```diff
@@ -1,10 +1,87 @@
...
+class ButtonDriver : public ButtonInterface
+{
  public:
-         virtual void Init()                = 0;
-         virtual bool Read()                = 0;
-         virtual bool Released()            = 0;
-         virtual bool Pressed()             = 0;
-         virtual void InvertButtonSignal(bool enable_invert_signal = true) = 0;
+       // Used for initialization of buttons
+       static constexpr uint8_t kFunction0 = 0b000;
```
kammce: Remove this in favor of using the function number within the Gpio class.

firmware/library/L1_Drivers/button_driver.hpp
```diff
@@ -1,10 +1,87 @@
...
+             gpio_(gpio),
+             pincon_(pincon),
+             gpio_pin_(port_, pin_),
+             pincon_pin_(Pin::CreateInactivePin())
+       {
+       }
+
+void Initialize() override {
+      pincon_pin_.SetMode(PinInterface::Mode::kInactive);
+      pincon_pin_.SetPinFunction(kFunction0);
```
kammce: Prefer that you use the internal gpio function variable: `Gpio::kGpioFunction`

firmware/library/L1_Drivers/button_driver.hpp
```diff
@@ -1,10 +1,87 @@
...
+      gpio_pin_.SetAsInput();
+}
+
+bool Read() override {
+    return (gpio_pin_.ReadPin()) ? true : false;
```
kammce: Rather than put this in the button driver, make a new function in the Gpio class and add a Read() function that returns bool.

firmware/library/L1_Drivers/button_driver.hpp
kammce: I think it would be logical to inherit the Gpio class rather than contain it.

firmware/library/L1_Drivers/button_driver.hpp
```diff
@@ -1,10 +1,87 @@
...
+#include <cstdint>
+
+#include "L0_LowLevel/LPC40xx.h"
+#include "L1_Drivers/gpio.hpp"
+#include "L1_Drivers/pin_configure.hpp"
+
+class ButtonInterface
+{
+ public:
+         virtual void Initialize()                        = 0;
```
kammce: run `clang-format -i <filename>` on this file.

firmware/library/L1_Drivers/button_driver.hpp
```diff
@@ -1,10 +1,87 @@
...
+void InvertButtonSignal(bool enable_invert_signal = true) override {
+  if (enable_invert_signal == false) {
+        pincon_pin_.SetAsActiveLow(false);
+  } else {
+        pincon_pin_.SetAsActiveLow(true);
+  }
+}
+
+~ButtonDriver() {}
+ private:
```
kammce: If you do inheritance for this, this section can all go away.

firmware/library/L1_Drivers/button_driver_test.cpp
```diff
@@ -0,0 +1,66 @@
...
+        constexpr uint32_t kMask          = 0b1 << kInvertPosition;
+       // Source: "UM10562 LPC408x/407x User manual" table 83 page 132
+       // Check that mapped pin P1.9's input inversion bit is set to 1
+       btn1_test.InvertButtonSignal(true);
+       CHECK(kMask == (local_iocon.P1_9 & kMask));
+       // Check that mapped pin P1.9's input inversion bit is set to 0
+       btn1_test.InvertButtonSignal(false);
+       CHECK(0 == (local_iocon.P1_9 & kMask));
+    }
+
```
kammce: Add tests for `Release()` and `Pressed()`. Use test by side-effect.

firmware/library/L1_Drivers/button_driver.hpp
Lawrence-Wan: There is a ReadPinBool() function in the Gpio class. Is this modification to the Gpio class still necessary?

kammce: Hmmm, it's not necessary since it exists, but I don't like that name anymore. I think, simply, `bool Read()` would be much better. Can you apply that change.

firmware/library/L3_HAL/button.hpp
```diff
@@ -0,0 +1,75 @@
...
+
+  void InvertButtonSignal(bool enable_invert_signal = true) override
+  {
+    SetAsActiveLow(enable_invert_signal);
+  }
+
+  ~Button() {}
+
+ private:
+  static bool was_pressed;
```
kammce: Why is this variable static? Does this mean we can only have 1 button? This variable would be referenced by ever Button class instance. I would think that this should be a variable that is owned by the object and not the class.

firmware/library/L3_HAL/button.hpp
```diff
@@ -0,0 +1,75 @@
...
+{
+ public:
+  constexpr Button(uint8_t port_num, uint8_t pin_num) : Gpio(port_num, pin_num)
+  {
+  }
+
+  void Initialize(void) override
+  {
+    SetMode(PinInterface::Mode::kPullDown);
+    SetAsInput();
```
kammce: Swap the two lines around. Set it as input then pullup.

firmware/library/L3_HAL/button_test.cpp
```diff
@@ -0,0 +1,84 @@
...
+  }
+  SECTION("Button Pressed and Button Released")
+  {
+    constexpr bool kExpectedResult = true;
+    constexpr uint8_t kPin9        = 9;
+
+    // Set P1_9 for button read == true
+    local_gpio_port[1].PIN |= (1 << kPin9);
+    // Simulate button being pressed.
+    test_subject.Released(true);
```
kammce: ... So I just realized how confusing this looks to pass a reset variable to the `Released` and `Pressed`, which is my bad, since I suggested it. I think having a `ResetState()`  function would be a cleaner alternative to this, so it pretty obvious what its doing.

PR #204, opened 2018-10-05
Implementation file for Uart and example code.

Jeremy-Chau:
Implementation file for Uart and example code.


Jeremy-Chau:
I need some help with the implementation of the look up table for the pin pairs. I don't understand why I can't assign the tx_pin_ and rx_pin_ variables as shown in the pull request like I did from before when I was using multiple tables for the pin pairs. For reference, below is how I assigned the tx and rx pins before.

        tx_pin_ = Pin(kTxChannel[mode_], kTxPortPin[mode_]);
        rx_pin_ = Pin(kRxChannel[mode_], kRxPortPin[mode_]);

kammce:
> Khalil,
> I can't figure out how to properly implement the Pin Pairs lookup table. I guess I don't quite understand how to use a look up table of function calls. I thought I could use the table in the same manner as I did before creating the table but the build errors state otherwise. For reference, below is how I was able to setup the pins before creating the table.
> 
> ```
>     tx_pin_ = Pin(kTxChannel[mode_], kTxPortPin[mode_]);
>     rx_pin_ = Pin(kRxChannel[mode_], kRxPortPin[mode_]);
> ```

With the way you are doing this, you can omit the `tx_pin_` and the `rx_pin_`, even though this should work. Simply use a map of Uart Pins and assign their pointers to the `tx_` and `rx_`  pointers.

kammce:
Even though coveralls says you decreased coverage by 2.6%, I will still merge you code in because it drops the binary size considerable (~3kB I think).

kammce:
Haven't been able to get the driver working yet. Will test more tomorrow.

kammce:
Figured out the issue. It mostly had to do with the fact that the UART0 pins were not correct. UART0 must be P0.2 and P0.3 since we need that for serial communication with host machine. Which means that the pin function bits were also off. After changing those and a few other things I got it to work.

See master's version of "Jeremy-Chau-UART" and make those changes to your repo. Ignore the changes to Hyperload though. I will make those changes after this is merged in.

Jeremy-Chau:
I believe I got all the changes in this updated commit. Please let me know if I missed any.

firmware/library/L1_Drivers/uart.hpp
```diff
@@ -1,12 +1,280 @@
...
+      continue;
+    }
+  }
+
+  uint8_t Receive(uint32_t timeout = 0x7FFFFFFF) override
+  {
+    uint8_t receiver         = 0;
+    uint64_t timeout_time    = Milliseconds() + timeout;
+    uint64_t current_time    = Milliseconds();
+    uint8_t timeout_receiver = uart_base_reg[channel_]->LSR & (1 << 0);
```
kammce: It isn't polling if you only check once.

Jeremy-Chau:
I should be finished with the code. Just can't figure out why the DLL register is empty from the testing file.

PR #205, opened 2018-10-05
Adding Apds9960 Implementation

DrLer0:
Added comments on how to use the low level gesture sensor driver.
To do: test file for this driver.

### DIFF
--- a/firmware/library/L3_HAL/apds_9960.hpp
+++ b/firmware/library/L3_HAL/apds_9960.hpp
@@ -1,20 +1,503 @@
+// Low level Gesture Driver for setting up APDS-9960 device
+// Usage:
+//  Apds9960 sensor;
+//  sensor.Initialize();
+//  sensor.EnableGesture();
+//  Apds9960::Gesture value = sensor.GetGesture();
 #pragma once
 
 #include <cstdint>
+#include <cstdlib>
+
+#include "L1_Drivers/i2c.hpp"
+#include "L2_Utilities/enum.hpp"
+#include "L2_Utilities/log.hpp"
+#include "L3_HAL/device_memory_map.hpp"
 
 class Apds9960Interface
 {
  public:
-  virtual void Initialize()                                    = 0;
-  virtual void EnableGesture()                                 = 0;
-  virtual void DisableGesture()                                = 0;
-  virtual void DecodeGestureData()                             = 0;
-  virtual uint8_t GetGesture()                                 = 0;
-  virtual uint8_t ReadMode()                                   = 0;
-  virtual uint8_t GetGestureEnterThreshhold()                  = 0;
-  virtual uint8_t GetGestureExitThreshhold()                   = 0;
-  virtual bool SetGestureEnterThresh(uint8_t threshhold_value) = 0;
-  virtual bool SetGestureExitThresh(uint8_t threshhold_value)  = 0;
-  virtual bool CheckIfGestureOccured()                         = 0;
-  virtual void ReadGestureFIFO()                               = 0;
+  template <device::Endian endianess, WriteFnt write, ReadFnt read>
+  SJ2_PACKED(struct)
+  MemoryMap_t  // APDS-9960
+  {
+    template <typename Int>
+    using Register = device::Register_t<Int, endianess, write, read>;
+
+    template <typename Int, size_t kLength>
+    using Array = device::Array_t<Int, kLength, write, read>;
+
+    device::Reserved_t<0x80> ram;
+    Register<uint8_t> enable;
+    Register<uint8_t> adc_integration_time;
+    device::Reserved_t<0x01> reserved0;
+    Register<uint8_t> wait_time_nongesture;
+    Register<uint16_t> als_interrupt_low_threshold;
+    Register<uint16_t> als_interrupt_high_threshold;
+    device::Reserved_t<0x01> reserved1;
+    Register<uint8_t> proximity_interrupt_low_threshold;
+    device::Reserved_t<0x01> reserved2;
+    Register<uint8_t> proximity_interrupt_high_threshold;
+    Register<uint8_t> interrupt_persistence_filters_nongesture;
+    Register<uint8_t> configuration_1;
+    Register<uint8_t> proximity_pulse_count_and_length;
+    Register<uint8_t> gain_control;
+    Register<uint8_t> configuration_2;
+    device::Reserved_t<0x01> reserved3;
+    Register<uint8_t> device_id;
+    Register<uint8_t> device_status;
+    Register<uint16_t> clear_channel_data;
+    Register<uint16_t> red_channel_data;
+    Register<uint16_t> green_channel_data;
+    Register<uint16_t> blue_channel_data;
+    Register<uint8_t> proximity_data;
+    Register<uint16_t> proximity_offset_up_right_down_left_photodiodes;
+    Register<uint8_t> configuration_3;
+    Register<uint8_t> gesture_proximity_enter_threshold;
+    Register<uint8_t> gesture_exit_threshold;
+    Register<uint16_t> gesture_configuration_1_2;
+    Register<uint16_t> gesture_up_down_offset;
+    Register<uint8_t> gesture_pulse_count_and_length;
+    Register<uint8_t> gesture_left_offset;
+    device::Reserved_t<0x01> reserved4;
+    Register<uint8_t> gesture_right_offset;
+    Register<uint16_t> gesture_configuration_3_4;
+    device::Reserved_t<0x02> reserved5;
+    Register<uint8_t> gesture_fifo_level;
+    Register<uint8_t> gesture_status;
+    device::Reserved_t<0x34> reserved6;
+    Register<uint8_t> force_interrupt;
+    Register<uint8_t> proximity_interrupt_clear;
+    Register<uint8_t> als_clear_channel_interrupt_clear;
+    Register<uint8_t> all_nongesture_interrupts_clear;
+    device::Reserved_t<0x14> reserved7;
+    union {
+      Array<uint8_t, 4> gesture_fifo_data;
+      struct
+      {
+        Register<uint8_t> up_value;
+        Register<uint8_t> down_value;
+        Register<uint8_t> left_value;
+        Register<uint8_t> right_value;
+      };
+    };
+  };
+  enum Direction : char
+  {
+    kNone  = 'x',
+    kUp    = 'u',
+    kDown  = 'd',
+    kLeft  = 'l',
+    kRight = 'r',
+    kFar   = 'f',
+    kNear  = 'n'
+  };
+  enum Gesture : char
+  {
+    kSwipeUP    = 'U',
+    kSwipeDOWN  = 'D',
+    kSwipeLEFT  = 'L',
+    kSwipeRIGHT = 'R',
+    kNEAR       = 'N',
+    kFAR        = 'F',
+    kError      = 'E'
+  };
+  enum Mode : uint8_t
+  {
+    kPOWERON = 0,
+    kALSENABLE,
+    kPROXIMITYDETECT,
+    kWAIT,
+    kALSINTERRUPT,
+    kPROXIMITYINTERRUPT,
+    kGESTURE
+  };
+
+  virtual bool FindDevice()                                       = 0;
+  virtual void Initialize()                                       = 0;
+  virtual bool SetMode(Mode mode, bool enable)                    = 0;
+  virtual void EnableGesture()                                    = 0;
+  virtual bool DisableGesture()                                   = 0;
+  virtual bool ReadGestureMode()                                  = 0;
+  virtual bool CheckIfGestureOccured()                            = 0;
+  virtual uint8_t GetGestureFIFOLevel()                           = 0;
+  virtual void ReadGestureFIFO(uint8_t * data, uint8_t fifolevel) = 0;
+  virtual void ProcessGestureData(uint8_t level)                  = 0;
+  virtual Gesture DecodeGestureData(uint8_t level)                = 0;
+  virtual Gesture GetGesture()                                    = 0;
+};
+
+class Apds9960 : public Apds9960Interface
+{
+ public:
+  static constexpr uint8_t kApds9960Address = 0x39;
+  // Avago-APDS-9960 : 32 x 4 byte FIFO = 128-bytes of data to store (page 33)
+  static constexpr uint8_t kMaxFifoSize = 128;
+  // User may change kMaxDataSize to adjust how many Direction Data points
+  // to be collected, then DecodeGestureData(uint8_t) can be overloaded to
+  // take care of new patterns of Direction points for different gestures
+  static constexpr uint8_t kMaxDataSize = 8;
+
+  // Defualt values for Gesture sensor initialization
+  enum DefaultValues : uint8_t
+  {
+    kNoMode                 = 0x00,
+    kAtime                  = 0xDB,
+    kWtime                  = 0xF6,
+    kPilt                   = 0x00,
+    kPiht                   = 0x32,
+    kPers                   = 0x11,
+    kConfig1                = 0x60,
+    kPpulse                 = 0x89,
+    kControl                = 0x05,
+    kConfig2                = 0x01,
+    kPoffsetUpRightDownLeft = 0x00,
+    kConfig3                = 0x00,
+    kGPEnTh                 = 0x14,
+    kGExTh                  = 0x64,
+    kGconfig1               = 0x40,
+    kGconfig2               = 0x66,
+    kGoffsetUpDown          = 0x00,
+    kGoffsetLeftRight       = 0x00,
+    kGPulse                 = 0xC5,
+    kGconfig3               = 0x00,
+    kGconfig4               = 0x02,
+    kAilt                   = 0xFF,
+    kAiht                   = 0x00
+  };
+
+  constexpr Apds9960()
+      : up_sensitivity_(-75),
+        down_sensitivity_(75),
+        left_sensitivity_(-50),
+        right_sensitivity_(50),
+        far_sensitivity_(750),
+        near_sensitivity_(1000),
+        far_count_(0),
+        near_count_(0),
+        index_(0)
+  {
+    i2c.Initialize();
+    for (int i = 0; i < kMaxDataSize; i++)
+    {
+      points_[i] = Direction::kNone;
+    }
+    for (int i = 0; i < kMaxFifoSize; i++)
+    {
+      gfifo_data_[i] = 0;
+    }
+  }
+  constexpr Apds9960(int8_t up_sensitivity, int8_t down_sensitivity,
+                     int8_t left_sensitivity, int8_t right_sensitivity,
+                     int16_t far_sensitivity, int16_t near_sensitivity)
+      : up_sensitivity_(up_sensitivity),
+        down_sensitivity_(down_sensitivity),
+        left_sensitivity_(left_sensitivity),
+        right_sensitivity_(right_sensitivity),
+        far_sensitivity_(far_sensitivity),
+        near_sensitivity_(near_sensitivity),
+        far_count_(0),
+        near_count_(0),
+        index_(0)
+  {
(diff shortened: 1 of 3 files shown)


firmware/library/L3_HAL/apds_9960.hpp
```diff
@@ -1,20 +1,444 @@
...
+  virtual void ProcessGestureData(uint8_t level) = 0;
+  virtual char DecodeGestureData() = 0;
+  virtual char GetGesture() = 0;
+};
+
+class Apds9960 : public Apds9960Interface
+{
+ public:
+  // Apds-9960 Registers
+  static constexpr uint8_t
```
kammce: I think should be within an `enum` but not an `enum class`. This wouldn't need to be strongly typed.

firmware/library/L3_HAL/apds_9960.hpp
```diff
@@ -1,20 +1,444 @@
...
+    return read_data;
+  }
+  void Initialize() override
+  {
+    I2cInterface::Status status;
+    i2c.Initialize();
+    status = i2c.Write(kApds9960Address, nullptr, 0);
+    if (status == I2cInterface::Status::kSuccess)
+    {
+      uint8_t initialization_sequence[56] =
```
kammce: This should be a `static constexpr` array. This will keep the array from being loaded during runtime. It should be set in stone at the time of compilation.

kammce: Might be good to take this out of this function and put it within the class scope.

firmware/library/L3_HAL/apds_9960.hpp
```diff
@@ -1,20 +1,444 @@
...
+        kGCONF2, kGconfig2,
+        kGoffsetU, kGoffsetUpDown,
+        kGoffsetD, kGoffsetUpDown,
+        kGoffsetL, kGoffsetLeftRight,
+        kGoffsetR, kGoffsetLeftRight,
+        kGPULSE, kGPulse,
+        kGCONF3, kConfig3,
+        kGCONF4, kGconfig4
+      };
+      DEBUG_PRINT("    Found device at address: 0x%02X", kApds9960Address);
```
kammce: I don't think this comment is too necessary. We should reserve comments for warnings. But! When we have a log level system, this would be a good comment to have for DEBUG level comments.

firmware/library/L3_HAL/apds_9960.hpp
```diff
@@ -1,20 +1,444 @@
...
+      i2c.Write(kApds9960Address, initialization_sequence,
+        sizeof(initialization_sequence));
+    }
+    else
+    {
+      DEBUG_PRINT("Error Initializing Gesture Sensor");
```
kammce: Even though it may seem silly, swap out the DEBUG_PRINT with an assert macro:  

    SJ2_ASSERT_WARNING(false, "Error ...");

firmware/library/L3_HAL/apds_9960.hpp
```diff
@@ -1,20 +1,444 @@
...
+    }
+  }
+  bool SetMode(uint8_t mode, bool enable) override
```
kammce: What is mode? Can you put a comment here to describe what the parameters do?

firmware/library/L3_HAL/apds_9960.hpp
```diff
@@ -1,20 +1,444 @@
...
+    WriteReg(kWTIME, 0xFF);
+    WriteReg(kPPULSE, 0x89);
+
+    WriteReg(kCONFIG2, 0xD0);
+
+    SetMode(0, 1);  // POWERON
+    SetMode(3, 1);  // WAIT
+    SetMode(2, 1);  // PROXIMITY
+    SetMode(6, 1);  // GESTURE
+    // kENABLE = 0100'1101
```
kammce: I'm not sure I understand what this means, I think each of these enable bits should be within their own `enum` for the class.

firmware/library/L3_HAL/apds_9960.hpp
```diff
@@ -1,20 +1,444 @@
...
+  }
+  bool DisableGesture() override
+  {
+    if (!SetMode(6, 0)) return false;
```
kammce: This 6 is another magic number, please give this a name, either in an `enum` or a `constexpr`

firmware/library/L3_HAL/apds_9960.hpp
```diff
@@ -1,20 +1,444 @@
...
+    return true;
+  }
+  uint8_t ReadGestureMode() override
+  {
+    uint8_t value = 0;
+    value = ReadReg(kGCONF4);
+    value &= ~(0xFE);
```
kammce: Prefer that you use the `&= (1 << 7)` then use the literal here. The shift and and is more intuitive.

kammce: I also noticed that you are clearing all the bits except for the MSB, is that intended?

firmware/library/L3_HAL/apds_9960.hpp
```diff
@@ -1,20 +1,444 @@
...
+    return value;
+  }
+  bool CheckIfGestureOccured() override
+  {
+    uint8_t gstatus = 0;
+    gstatus = ReadReg(kGSTATUS);
+
+    gstatus &= 0b00000001;
```
kammce: I think an `&= (1 << 0)` would do fine here.

firmware/library/L3_HAL/apds_9960.hpp
```diff
@@ -1,20 +1,444 @@
...
+
+    if (gstatus == 1)
```
kammce: Rather than doing masking, then doing an if/else statement with seperate returns, you could simply do the following:

    return (gstatus & 0b1);

Its a lot cleaner and to the point.

firmware/library/L3_HAL/apds_9960.hpp
```diff
@@ -1,20 +1,444 @@
...
+    {
+    return true;
+    } else
+    {
+    return false;
+    }
+  }
+  void ReadGestureFIFO(uint8_t * data, uint8_t fifolvl) override
```
kammce: Spell out `fifo_level`, rather than truncating variables names. Try not to abbreviate.

firmware/library/L3_HAL/apds_9960.hpp
```diff
@@ -1,20 +1,444 @@
...
+  {
+    uint8_t fifo_registers = kGFifoU;
```
kammce: This should be `constexpr`

firmware/library/L3_HAL/apds_9960.hpp
```diff
@@ -1,20 +1,444 @@
...
+    i2c.WriteThenRead(kApds9960Address, &fifo_registers, 1, data, (4*fifolvl));
+  }
+  uint8_t GetGestureFIFOLvl() override
+  {
+    uint8_t value = 0;
+    value = ReadReg(kGSTATUS);
+    value &= 0b00000010;  // GFOV Mask
+    if (value)
```
kammce: You can combine this and the above lines ->
```
if(value & 0b10)
```

firmware/library/L3_HAL/apds_9960.hpp
```diff
@@ -1,20 +1,444 @@
...
+    {
+      uint8_t level = 0;
+      level = ReadReg(kGFLVL);
+
+      uint8_t throw_away[128];
```
kammce: Why 128? If this number means something make it a constant.

firmware/library/L3_HAL/apds_9960.hpp
```diff
@@ -1,20 +1,444 @@
...
+    return true;
+    } else
+    {
+    return false;
+    }
+  }
+  void ReadGestureFIFO(uint8_t * data, uint8_t fifolvl) override
+  {
+    uint8_t fifo_registers = kGFifoU;
+    i2c.WriteThenRead(kApds9960Address, &fifo_registers, 1, data, (4*fifolvl));
```
kammce: Why the number 4? Is this the size of the data? Is it the size `int32_t`? If so, use `sizeof(int32_t)` instead, to give some context for the value.

firmware/library/L3_HAL/apds_9960.hpp
```diff
@@ -1,20 +1,444 @@
...
+      uint8_t throw_away[128];
+      while (level != 0)
+      {
+        ReadGestureFIFO(throw_away, level);
+        level = ReadReg(kGFLVL);
+      }
+    }
+    return ReadReg(kGFLVL);
+  }
+  void ProcessGestureData(uint8_t level) override
```
kammce: Can you comment what each stage of this algorithm does. Its pretty cryptic otherwise.

firmware/library/L3_HAL/apds_9960.hpp
```diff
@@ -1,20 +1,444 @@
...
+  char DecodeGestureData() override
+  {
+    for (int i = 0; i < 8; i++)
+    {
+      switch (points_[i])
+      {
+      case Direction::kUp:
+          if (points_[i+1] == Direction::kDown)
+          {
+              // DEBUG_PRINT("Down Gesture\n");
```
kammce: Remove all debug comments from production code.

firmware/library/L3_HAL/apds_9960.hpp
```diff
@@ -1,20 +1,444 @@
...
+              return static_cast<char>(Gesture::kSwipeDOWN);
```
kammce: Prefer not to use multiple return statements. Have a singular Gesture type variable above and set it based on which case has been set. Then return it at the end of the function, with the necessary cast.

firmware/library/L3_HAL/apds_9960.hpp
```diff
@@ -1,20 +1,444 @@
...
+                  + gfifo_data_[i+3]) < near_sensitivity_ )  // Far
+      {
+          far_count_++;
+      } else
+      {
+          // do nothing
+      }
+    }
+  }
+  char DecodeGestureData() override
```
kammce: Why return `char`? Why not return `Gesture` type?

firmware/library/L3_HAL/apds_9960.hpp
```diff
@@ -1,20 +1,444 @@
...
+          } else
+          {
+            // do nothing
+          }
+          break;
+      }
+    }
+    return static_cast<char>(Gesture::kERROR);
+  }
+  char GetGesture() override
```
kammce: This too should return Gesture.

firmware/library/L3_HAL/apds_9960.hpp
```diff
@@ -1,20 +1,444 @@
...
+  {
+    if (CheckIfGestureOccured())
+    {
+    // check if gesture ended
```
kammce: Fix indentation. Run `clang-format` on your code please.

firmware/library/L3_HAL/apds_9960.hpp
```diff
@@ -1,20 +1,444 @@
...
+      if (ReadGestureMode() == 0)
+      {
+        uint8_t level = 0;
+        // print out raw gesture data
+        DEBUG_PRINT("start gesture:\n");
+        DEBUG_PRINT("flvl: %i\n", GetGestureFIFOLvl());
```
kammce: Remove debug statements from productions code.

firmware/library/L3_HAL/apds_9960.hpp
```diff
@@ -1,20 +1,444 @@
...
+        level = GetGestureFIFOLvl();
+
+        ReadGestureFIFO(gfifo_data_, (4*level));
+
+        for (int i = 0; i < (4*level); i+=4)
```
kammce: Remove debug statements from productions code.

firmware/library/L3_HAL/apds_9960.hpp
```diff
@@ -1,20 +1,444 @@
...
+        {
+          DEBUG_PRINT("%i, %i, %i, %i\n", gfifo_data_[i], gfifo_data_[i+1] ,
+            gfifo_data_[i+2], gfifo_data_[i+3]);
+        }
+
+        DEBUG_PRINT("end gesture:\n");
```
kammce: Remove debug statements from productions code.

firmware/library/L3_HAL/apds_9960.hpp
```diff
@@ -1,20 +1,444 @@
...
+        // print out raw gesture data
+        DEBUG_PRINT("start gesture:\n");
+        DEBUG_PRINT("flvl: %i\n", GetGestureFIFOLvl());
+        level = GetGestureFIFOLvl();
+
+        ReadGestureFIFO(gfifo_data_, (4*level));
+
+        for (int i = 0; i < (4*level); i+=4)
+        {
+          DEBUG_PRINT("%i, %i, %i, %i\n", gfifo_data_[i], gfifo_data_[i+1] ,
```
kammce: This would be better suited for your Gesture sensor example, then in your driver.

firmware/library/L3_HAL/apds_9960.hpp
```diff
@@ -1,20 +1,444 @@
...
+        }
+
+        ProcessGestureData(level);
+
+        return DecodeGestureData();
+      }
+    }
+    return static_cast<char>(Gesture::kERROR);
+  }
+  explicit Apds9960(int8_t up_sensitivity = -75,
```
kammce: Make this constexpr

firmware/library/L3_HAL/apds_9960.hpp
```diff
@@ -1,20 +1,444 @@
...
+ private:
+  static constexpr uint8_t kApds9960Address = 0x39;
+  int8_t up_sensitivity_;
+  int8_t down_sensitivity_;
+  int8_t left_sensitivity_;
+  int8_t right_sensitivity_;
+  uint32_t far_sensitivity_;
+  uint32_t near_sensitivity_;
+
+  uint8_t gfifo_data_[128] = {};
```
kammce: Again, why 128?

firmware/library/L3_HAL/apds_9960.hpp
```diff
@@ -1,20 +1,518 @@
...
+  virtual void ProcessGestureData(uint8_t level)                  = 0;
+  virtual Gesture DecodeGestureData(uint8_t level)                = 0;
+  virtual Gesture GetGesture()                                    = 0;
+};
+
+class Apds9960 : public Apds9960Interface
+{
+ public:
+  // defualt values for sensor initialization
+  static constexpr uint8_t kNoMode = 0x00, kAtime = 0xDB, kWtime = 0xF6,
```
kammce: I feel that an enumeration would look a bit cleaner here.

firmware/library/L3_HAL/apds_9960.hpp
```diff
@@ -1,20 +1,518 @@
...
+                           kConfig3 = 0x00, kGPEnTh = 0x14, kGExTh = 0x64,
+                           kGconfig1 = 0x40, kGconfig2 = 0x66,
+                           kGoffsetUpDown = 0, kGoffsetLeftRight = 0,
+                           kGPulse = 0xC5, kGconfig3 = 0x00, kGconfig4 = 0x02;
+
+  bool FindDevice() override
+  {
+    i2c.Initialize();
+    uint8_t device_id = gesture.memory.device_id;
+    if (device_id == 0xAB)
```
kammce: You should be able to do this in 1 line rather than having to store this in a variable first. You don't lose any efficiency, but it would look cleaner and the code would be shorter.

firmware/library/L3_HAL/apds_9960.hpp
```diff
@@ -1,20 +1,518 @@
...
+    {
+      return true;
```
kammce: Prefer to only return once in a function. So you can have a, `bool device_found = false` and set that to true when the device_id comes back as 0xAB.

firmware/library/L3_HAL/apds_9960.hpp
```diff
@@ -1,20 +1,518 @@
...
+                 down_sensitivity_)  // Down
+        {
+          points_[index_] = Direction::kDown;
+          index_++;
+        }
+      }
+      else
+      {
+        // If a photodiode has a smaller value,
+        //  then that photodiode was covered and movement detected
```
kammce: Nit: remove the extra space.

firmware/library/L3_HAL/apds_9960.hpp
```diff
@@ -1,20 +1,518 @@
...
+        near_count_++;
+      }
+      // If sum is < far_sensitivity_ threshold,
+      //    then dataset represents a far action
+      else if ((gfifo_data_[i] + gfifo_data_[i + 1] + gfifo_data_[i + 2] +
+                gfifo_data_[i + 3]) < far_sensitivity_)  // Far
+      {
+        far_count_++;
+      }
+      else
```
kammce: You can remove the else if it doesn't do anything.

firmware/library/L3_HAL/apds_9960.hpp
```diff
@@ -1,20 +1,518 @@
...
+      {
+        // do nothing
+      }
+    }
+  }
+  Gesture DecodeGestureData(uint8_t level) override
+  {
+    Gesture result = kERROR;
```
kammce:
```suggestion
    Gesture result = kError;
```

firmware/library/L3_HAL/apds_9960.hpp
```diff
@@ -1,20 +1,518 @@
...
+    for (int i = 0; i < kMaxDataSize; i++)
+    {
+      points_[i] = Direction::kNone;
+    }
+    for (int i = 0; i < kMaxFifoSize; i++)
+    {
+      gfifo_data_[i] = 0;
+    }
+  }
+  ~Apds9960() {}
```
kammce: You can remove this.

firmware/library/L3_HAL/apds_9960.hpp
```diff
@@ -1,20 +1,518 @@
...
+
+ private:
+  // Max FIFO size is represented by the max 4x32 FIFO for
+  // Up/Down/Left/Right gesture data
+  static constexpr uint8_t kMaxFifoSize = 128;
+  static constexpr uint8_t kMaxDataSize = 8;
```
kammce: Do you think the user would ever want or need to change these values? If so, we should put this in the config.hpp file. If not, then put a comment saying to keep these values as they are and where you found/calculated them.

kammce:
Also, please update your branch.

firmware/library/L3_HAL/apds_9960.hpp
```diff
@@ -1,20 +1,518 @@
...
+
+  device::Reserved_t<0x80> ram;
+  Register<uint8_t> enable;
+  Register<uint8_t> adc_integration_time;
+  device::Reserved_t<0x01> reserved0;
+  Register<uint8_t> wait_time_nongesture;
+  Register<uint8_t> als_interrupt_low_threshold_low_byte;
+  Register<uint8_t> als_interrupt_low_threshold_high_byte;
+  Register<uint8_t> als_interrupt_high_threshold_low_byte;
+  Register<uint8_t> als_interrupt_high_threshold_high_byte;
```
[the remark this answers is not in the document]
DrLer0: I've combined the respective low-high bytes to make a 16-bit number for a low and high threshold value. They're distinct values to be compared to CDATA (page 21)

firmware/library/L3_HAL/apds_9960.hpp
```diff
@@ -1,20 +1,518 @@
...
+    {
+      return false;
+    }
+    return true;
+  }
+  bool ReadGestureMode() override
+  {
+    uint8_t value = 0;
+    value         = gesture.memory.gesture_configuration_four;
+    return (value & 0b1);
```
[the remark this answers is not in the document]
DrLer0: I had trouble doing this as one line and combining gesture_configuration_4 and gesture_configuration_3 into a uint16_t register. The gesture state machine exited too early (i.e. data stopped collecting for a moment, and then the gesture state machine restarted)

PR #211, opened 2018-10-10
Added buzzer unit test; added getters for frequency and volume

Hashah1:


firmware/library/L3_HAL/test/buzzer_test.cpp
```diff
@@ -0,0 +1,56 @@
...
+    }
+
+    SECTION("Check Beep")
+    {
+        constexpr uint32_t kFrequency = 0;
+        constexpr float kVolume = 0.0;
+        test1.Beep(kFrequency, kVolume);
+
+        float vol_error = kVolume - test1.GetVolume();
+        CHECK((-0.1f <= vol_error && vol_error <= 0.1f) == true);
```
kammce:
```suggestion
        CHECK(-0.1f <= vol_error && vol_error <= 0.1f);
```
You need to remove the == true from this statement. CHECK evaluates the condition within the () and checks if it is true already. No need to do the == true.

firmware/library/L3_HAL/test/buzzer_test.cpp
```diff
@@ -0,0 +1,56 @@
...
+    }
+
+    SECTION("Check Stop")
+    {
+        constexpr uint32_t kFrequency = 500;
+        constexpr float kVolume = 0.0;
+
+        test1.Stop();
+        float vol_error = kVolume - test1.GetVolume();
+        CHECK((-0.1f <= vol_error && vol_error <= 0.1f) == true);
```
kammce:
```suggestion
        CHECK(-0.1f <= vol_error && vol_error <= 0.1f);
```
Same as the above comment.

PR #212, opened 2018-10-10
Fixed an issue with static allocation of Tasks

HuangDave:
When declaring cstring pointer arrays, the string length should be specified to
prevent memory issues or hard fault errors.

Added a nullptr check to Delete() and RemoveTask() to allow tasks to be removed
even if the scheduler has not been started.

### DIFF
--- a/firmware/examples/TaskScheduler/source/main.cpp
+++ b/firmware/examples/TaskScheduler/source/main.cpp
@@ -2,28 +2,40 @@
 
 #include "L2_Utilities/log.hpp"
 #include "L4_Application/task.hpp"
+#include "semphr.h"
 
-class PrinterTask : public rtos::Task<1024>
+class PrinterTask : public rtos::Task<256>
 {
  public:
-  PrinterTask(const char * task_name, const char * message)
+  constexpr PrinterTask(const char * task_name, const char * message)
       : Task(task_name, rtos::Priority::kMedium),
         message_(message),
-        run_count_(0){};
+        run_count_(0)
+  {
+    print_mutex = {0};
+  };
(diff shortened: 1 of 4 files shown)


firmware/examples/TaskScheduler/source/main.cpp
```diff
@@ -39,15 +39,27 @@ class PrinterTask : public rtos::Task<1024>
...
-    // TODO(#210): Should remove the use of new and use static allocation for
-    //             constructing tasks when the issue is fixed.
-    PrinterTask * printer_one =
-        new PrinterTask("Printer A", "I am a printer, I am faster");
-    printer_one->SetDelayTime(500);
-    PrinterTask * printer_two =
-        new PrinterTask("Printer B", "I am also a printer");
-    printer_two->SetDelayTime(1000);
+
+    PrinterTask printer_one(task_names[0], messages[0]);
```
kammce: The developer should be able to put a string literal here, without needing to declare it as a C-style array above.

firmware/examples/TaskScheduler/source/main.cpp
```diff
@@ -32,30 +44,27 @@ class PrinterTask : public rtos::Task<1024>
   };
 
  private:
+  inline static SemaphoreHandle_t print_mutex;
   const char * message_;
   uint32_t run_count_;
 };
 
+PrinterTask printer_one("Printer A", "I am a printer, I am faster");
```
kammce: Can you put a comment here to tell the user, that they must be declared globally or be allocated on the heap using the `new` operator for this to work. Otherwise, the system will crash.

PR #214, opened 2018-10-12
Adding ESP8266 Interface

PricelessAntonio:


firmware/library/L1_Drivers/esp8266.hpp
```diff
@@ -0,0 +1,37 @@
...
+    virtual void ResetModule()                                   = 0;
+    virtual void DeepSleep(uint32_t time)                        = 0;
+    virtual void GetModuleVersion()                              = 0;
+    virtual void GetNetworkConnection()                          = 0;
+    virtual bool ConnectToAccessPoint(
+        const char ssid[], const char password[])                = 0;
+    virtual void DisconnectFromAccessPoint()                     = 0;
+    virtual bool IsConnectedToAccessPoint()                      = 0;
+    virtual bool ConnectToRemoteIp(
+        const char remote_ip[], const char port[],
```
kammce: I'd rather this say hostname rather than ip address. Since either one can be used, I think we should favor a variable name that fits closer to what people are going to remember.

I can remember google.com, yahoo.com, sjsu.edu, but I cannot remember or track their ip addresses.

firmware/library/config.hpp
```diff
@@ -91,4 +92,10 @@ SJ2_DECLARE_CONSTANT(DEBUG_PRINT_ENABLED, bool, kDebugPrintEnabled);
 #endif  // !defined SJ2_TASK_SCHEDULER_SIZE
 SJ2_DECLARE_CONSTANT(TASK_SCHEDULER_SIZE, uint8_t, kTaskSchedulerSize);
 
+// Used to set the reciever buffer size of the ESP8266 driver
+#if !defined SJ2_ESP8266_BUFFER_SIZE
+#define SJ2_ESP8266_BUFFER_SIZE 500
```
kammce: Would recommend 512. Its a common base 2 number.

firmware/library/L1_Drivers/esp8266.hpp
```diff
@@ -0,0 +1,37 @@
...
+        ConnectionType type = ConnectionType::kTcp)              = 0;
+    virtual void DisconnectFromRemoteIp()                        = 0;
+    virtual bool IsConnectedToRemoteIp()                         = 0;
+    virtual bool SendGetRequest(
+        const char url[], char buffer[] = default_buffer.data()) = 0;
+    virtual bool SendPostRequest(
```
kammce: This and the above class should take a `size_t length` of the buffer. This needs to be known.

PR #266, opened 2018-11-27
Adding Power On System Test

delwin-lei:
Testing Gpio, LEDs, Accelerometer, Gesture, Temperature sensor


delwin-lei:
I will fix current Travis errors and the folder name by tomorrow.

firmware/examples/Post/source/main.cpp
```diff
@@ -0,0 +1,287 @@
...
+
+  LOG_INFO("Initializing OLED Hardware Test...");
+  display.Initialize();
+
+  LOG_INFO("Clearing Screen...");
+  display.Clear();
+
+  LOG_INFO("Clearing Screen finished. Waiting 1 second...");
+
+  Delay(1000);
```
kammce: This is actually not required. Should work if you remove this.

You can remove the "Waiting 1 second..." part of hte message above.

firmware/examples/Post/source/main.cpp
```diff
@@ -0,0 +1,287 @@
...
+  uint8_t array[5];
+  cs.SetLow();
+  Delay(1);
+  // Read Manufacturer ID
+  ssp2.Transfer(0x9F);
+  array[0] = static_cast<uint8_t>(ssp2.Transfer(0x00));
+  array[1] = static_cast<uint8_t>(ssp2.Transfer(0x00));
+  array[2] = static_cast<uint8_t>(ssp2.Transfer(0x00));
+  array[3] = static_cast<uint8_t>(ssp2.Transfer(0x00));
+  LOG_INFO("Returned 0x%02X 0x%02X 0x%02X 0x%02X", array[0], array[1], array[2],
```
kammce: Wait, wait, wait. What are you testing here? What if the bytes coming back area all zeros or all `0xff`? You need to check the values to verify that the test actually worked. Remember that the people in the factory will not be reading the output from the console. And if we made them do that, that would be very confusing to have to look for a specific pattern of bytes, when we could just simply check them in the code.

firmware/examples/Post/source/main.cpp
```diff
@@ -0,0 +1,287 @@
...
+  // Creates a status variable for I2C transactions
+  Status status;
+  printf("++++++++++++++++++++++++++++++++++++++\n\n");
+  printf("Beginning I2c scan...\n\n");
+  for (uint8_t address = kFirstI2cAddress; address < kLastI2cAddress; address++)
+  {
+    status = i2c.Write(address, nullptr, 0);
+    if (status == Status::kSuccess)
+    {
+      LOG_INFO("    Found device at address: 0x%02X", address);
```
kammce: Same issue here. Don't worry about printing stuff to console. You need to make sure that the addresses you find, are the correct addresses. This four loop should basically be searching for the 3 I2C addresses we expect to come back. You can use the log statements here to get what those addresses will be exactly, but you cannot omit actually checking them.

firmware/examples/Post/source/main.cpp
```diff
@@ -0,0 +1,287 @@
...
+  Delay(1000);
+}
+
+bool GestureSensorTest()
+{
+  // Gesture Sensor Test
+  printf("++++++++++++++++++++++++++++++++++++++\n\n");
+  printf("Starting Gesture Sensor Test...\n\n");
+  uint8_t kApds9960IdRegisterAddress = 0x92;
+  // uint8_t control_sequence[] = { 0x8F, 0x00 };
```
kammce: Delete any unused code. No need to clutter up the factory test.

firmware/examples/Post/source/main.cpp
```diff
@@ -0,0 +1,287 @@
...
+    { 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0,
+      0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0 },
+    { 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0,
+      0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 1, 1, 0, 0 }
+  };
+
+  printf("++++++++++++++++++++++++++++++++++++++\n\n");
+  printf("Begin Gpio test...\n\n");
+
+  uint8_t num_ports                   = 5;
```
kammce: This should be constexpr and placed above the array. Then the values should be used to setup the fields for the 2D array. No need to have to write the number 5 or 32 twice in this section of code.

firmware/examples/Post/source/main.cpp
```diff
@@ -0,0 +1,287 @@
...
+  bool temp_pass          = TemperatureSensorTest();
+  bool final_result       = gesture_pass && accelerometer_pass && temp_pass;
+  return final_result;
+}
+
+bool GpioFunctionalityTest()
+{
+  // P0_2 and P0_3 do not need to be tested.
+  // P0_29 and P0_30 do not need to be tested (lack of pullup/pulldown)
+  const bool LQFP80[5][32] = {
```
kammce: It might be more readable to use this array designator syntax for the 2D arrays. See: https://en.cppreference.com/w/c/language/array_initialization

firmware/examples/Post/source/main.cpp
```diff
@@ -0,0 +1,287 @@
...
+        else
+        {
+          LOG_INFO("Pin[%u][%u] Value:%i", port, pin, test_pin.Read());
+          gpio_test[port][pin] = false;
+          if (LQFP80[port][pin] == true && gpio_test[port][pin] == false)
+          {
+            LOG_INFO("Gpio[%u][%u] failed at stage 2.", port, pin);
+          }
+        }
+        gpio_total++;
```
kammce: The Gpio should be reverted back to their original state after this test (Input & pull-up).

firmware/examples/Post/source/main.cpp
```diff
@@ -0,0 +1,287 @@
...
+        else
+        {
+          LOG_INFO("Pin[%u][%u] Value:%i", port, pin, test_pin.Read());
+          gpio_test[port][pin] = false;
+          if (LQFP80[port][pin] == true && gpio_test[port][pin] == false)
+          {
+            LOG_INFO("Gpio[%u][%u] failed at stage 1.", port, pin);
+          }
+        }
+        test_pin.SetHigh();
```
kammce: You aren't doing the pull-up/pull-down test anywhere in this code. Where is it?

firmware/examples/Post/source/main.cpp
```diff
@@ -0,0 +1,287 @@
...
+          {
+            LOG_INFO("Gpio[%u][%u] failed at stage 2.", port, pin);
+          }
+        }
+        gpio_total++;
+      }
+    }
+  }
+  LOG_INFO("Number of Gpio pins that passed: %u/%u", gpio_passed, gpio_total);
+  return gpio_passed / gpio_total;
```
kammce:
```suggestion
  return gpio_passed == gpio_total;
```
Why are you doing it this way? Why not just check that they are equal? Why use divide? This could, potentially result in some problems if there are rounding errors during this calculation.

firmware/examples/Post/source/main.cpp
```diff
@@ -0,0 +1,287 @@
...
+  i2c.Write(kGestureAddress, initialization_sequence,
+            sizeof(initialization_sequence));
+
+  i2c.WriteThenRead(kGestureAddress, &kApds9960IdRegisterAddress, 1,
+                    &kApds9960IdRegisterAddress, 1);
+
+  LOG_INFO("Gesture ID = 0x%02X", kApds9960IdRegisterAddress);
+  LOG_INFO("End of Gesture Sensor Test...");
+
+  return true;
```
kammce: Again... nothing is being tested here.

firmware/examples/Post/source/main.cpp
```diff
@@ -0,0 +1,287 @@
...
+  i2c.Write(kAccelerometerAddress, initialization_sequence,
+            sizeof(initialization_sequence));
+
+  i2c.WriteThenRead(kAccelerometerAddress, &kMMAIdRegisterAddress, 1,
+                    &kMMAIdRegisterAddress, 1);
+
+  LOG_INFO("Accelerometer ID = 0x%02X", kMMAIdRegisterAddress);
+  LOG_INFO("End of Accelerometer Test...");
+
+  return true;
```
kammce: Again nothing is being tested here.

firmware/examples/Post/source/main.cpp
```diff
@@ -0,0 +1,287 @@
...
+  LOG_INFO("temperature_data is %02X%02X, %04X", temperature_ms, temperature_ls,
+           temperature_data);
+  // Compute the actual temperature in Celsius
+  int actual_temp = 55 + ((temperature_data - 16384) / 160);
+  LOG_INFO("Temperature is: %u", actual_temp);
+  LOG_INFO("End of temperature test...");
+  if (-40 < actual_temp && actual_temp < 125)
+    return true;
+  else
+    return false;
```
kammce: Finally some tests.  But also, use curly braces always.

firmware/examples/FactoryTest/source/main.cpp
```diff
@@ -0,0 +1,323 @@
...
+#include "L1_Drivers/gpio.hpp"
+#include "L1_Drivers/i2c.hpp"
+#include "L1_Drivers/ssp.hpp"
+#include "L2_Utilities/log.hpp"
+#include "L2_Utilities/status.hpp"
+#include "L3_HAL/onboard_led.hpp"
+#include "L3_HAL/ssd1306.hpp"
+
+constexpr uint8_t kFirstI2cAddress = 0x08;
+constexpr uint8_t kLastI2cAddress  = 0x78;
```
kammce: This is only used in 1 function. Put them in there.

firmware/examples/FactoryTest/source/main.cpp
```diff
@@ -0,0 +1,323 @@
...
+      gesture_pass && accelerometer_pass && temp_pass && i2c_pass;
+  return final_result;
+}
+
+bool GpioFunctionalityTest()
+{
+  // P0_2 and P0_3 do not need to be tested.
+  // P0_29 and P0_30 do not need to be tested (lack of pullup/pulldown)
+  constexpr uint8_t num_ports = 5;
+  constexpr uint8_t num_pins  = 32;
```
kammce:
```suggestion
  constexpr uint8_t kNumPins  = 32;
```
Please follow google c++ standards.

firmware/examples/FactoryTest/source/main.cpp
```diff
@@ -0,0 +1,323 @@
...
+    { 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0,
+      0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0 },
+    { 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0,
+      0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 1, 1, 0, 0 }
+  };
+
+  printf("++++++++++++++++++++++++++++++++++++++\n\n");
+  printf("Begin Gpio test...\n\n");
+
+  uint8_t gpio_total                  = 0;
```
kammce:
```suggestion
  constexpr uint8_t kGpioTotal                  = GetTotalOfGpiosToTest(kLQFP80);
```
This should be constexpr. You can make kLQFP80 a constexpr array, pass that to a constexpr function, that will return the total by checking all of the 1s in the bitmap.

firmware/examples/FactoryTest/source/main.cpp
```diff
@@ -0,0 +1,323 @@
...
+}
+
+bool GpioFunctionalityTest()
+{
+  // P0_2 and P0_3 do not need to be tested.
+  // P0_29 and P0_30 do not need to be tested (lack of pullup/pulldown)
+  constexpr uint8_t num_ports = 5;
+  constexpr uint8_t num_pins  = 32;
+
+  const bool kLqfP80[num_ports][num_pins] = {
```
kammce:
```suggestion
  const bool kLPC4078[num_ports][num_pins] = {
```

I think this would be a better name. Because kLqfp80 is a package size, but it doesn't tell you what part you are working with. The LPC40xx series has a bunch of size chips, but the LPC4078 is specifically the package LQFP80. Using the actual part number, I think, would make more sense here.

firmware/examples/FactoryTest/source/main.cpp
```diff
@@ -0,0 +1,323 @@
...
+  for (uint8_t port = 0; port < num_ports; port++)
+  {
+    for (uint8_t pin = 0; pin < num_pins; pin++)
+    {
+      if (kLqfP80[port][pin])
+      {
+        Gpio test_pin(port, pin);
+        test_pin.SetAsOutput();
+        test_pin.SetLow();
+        if (test_pin.Read() == false)
```
kammce:
```suggestion
        if (!test_pin.Read())
```

firmware/examples/FactoryTest/source/main.cpp
```diff
@@ -0,0 +1,323 @@
...
+        {
+          gpio_test[port][pin] = true;
+        }
+        else
+        {
+          LOG_INFO("Pin[%u][%u] Value:%i", port, pin, test_pin.Read());
+          gpio_test[port][pin] = false;
+          if (kLqfP80[port][pin] == true && gpio_test[port][pin] == false)
```
kammce: Why are you checking this again? This block of code can never be reached if `kLqfP80[port][pin] == false` due to the first `if` statement.

kammce: `gpio_test[port][pin] == false` is always false, because the line directly above it sets it to false. Why is it being checked here?

firmware/examples/FactoryTest/source/main.cpp
```diff
@@ -0,0 +1,323 @@
...
+          {
+            LOG_INFO("Gpio[%u][%u] failed at stage 1.", port, pin);
```
kammce: You should simply combine this and the above `LOG_INFO` statements.

firmware/examples/FactoryTest/source/main.cpp
kammce: Use a helper function like `bool SetPinCanBeSetLow()` and `bool `SetPinCanBeSetHigh()`. Its kinda hard to see whats happening with so much code in one place. Spread it out just a bit.

firmware/examples/FactoryTest/source/main.cpp
```diff
@@ -0,0 +1,323 @@
...
+          gpio_passed++;
+        }
+        else
+        {
+          LOG_INFO("Pin[%u][%u] Value:%i", port, pin, test_pin.Read());
+          gpio_test[port][pin] = false;
+          if (kLqfP80[port][pin] == true && gpio_test[port][pin] == false)
+          {
+            LOG_INFO("Gpio[%u][%u] failed at stage 2.", port, pin);
+          }
```
kammce: Please test pull-up and pull-downs as well here.

firmware/examples/FactoryTest/source/main.cpp
```diff
@@ -0,0 +1,323 @@
...
+{
+  // Turn on all LEDs if all tests pass else none.
+  // LED Test
+  OnBoardLed leds;
+  leds.Initialize();
+  if (gpio_test && i2c_test && flash_test)
+  {
+    leds.SetAll(0x0F);
+  }
+  LOG_INFO("Test Results:\n\tGPIO:%d\n\tI2C:%d\n\tFlash:%d\n\t", gpio_test,
```
kammce: This is a bit hard to read. I would suggest simply multiple printf statements here.

firmware/examples/FactoryTest/source/main.cpp
```diff
@@ -0,0 +1,323 @@
...
+  // uint8_t control_sequence[] = { 0x8F, 0x00 };
+  // i2c.WriteThenRead(kGestureAddress, &control_sequence[0], 1,
+  //                   &control_sequence[0], 1);
+  // LOG_INFO("Control 1 = 0x%02X", control_sequence[0]);
+  // control_sequence[0] = 0x90;
+  // i2c.WriteThenRead(kGestureAddress, &control_sequence[0], 1,
+  //                   &control_sequence[0], 1);
+  // LOG_INFO("Control 2 = 0x%02X", control_sequence[0]);
+  // control_sequence[0] = 0x9F;
+  // i2c.Write(kGestureAddress, control_sequence, sizeof(control_sequence));
```
kammce: Delete this commented out code.

firmware/examples/FactoryTest/source/main.cpp
```diff
@@ -0,0 +1,323 @@
...
+  LOG_INFO("End of temperature test...");
+  if (-40 < actual_temp && actual_temp < 125)
+  {
+    result &= true;
+  }
+
+  return result;
+}
+
+bool CheckDeviceId(uint8_t actual_id, uint8_t expected_id)
```
kammce: This function is pretty useless. It would actually be more useful to simply just do the equation in the code where this is called. What this should do is the following:

```
bool CheckDeviceId(I2c & i2c_peripheral, uint8_t device_address, uint8_t id_register_address, uint8_t  expected_id)
{
   // Do I2c transaction to figure stuff out here
}
```

If you notice, all of your id check calls have the same pattern, with a few values made differently. So you can make a function that actually does the i2c transaction for you and simply pass it a few bits of info about the device.

firmware/examples/FactoryTest/source/main.cpp
```diff
@@ -0,0 +1,323 @@
...
+  uint8_t kTemperatureIdAddress = 0xC0;
+  uint8_t temperature_id        = 0x00;
+  bool result                   = false;
+
+  i2c.WriteThenRead(kTemperatureAddress, &kTemperatureIdAddress, 1,
+                    &temperature_id, 1);
+  LOG_INFO("ID:%02X", temperature_id);
+  result = CheckDeviceId(temperature_id, 0x14);
+
+  uint8_t control_sequence[] = { 0xC4, 0x00, 0xC9, 0x01 };
```
kammce: Create a helper function for calculating temperature. Pass an i2c reference to it.

firmware/examples/FactoryTest/source/main.cpp
```diff
@@ -0,0 +1,323 @@
...
+  temperature_data = ((temperature_ms & 0x7F) << 8) | temperature_ls;
+  LOG_INFO("temperature_data is %02X%02X, %04X", temperature_ms, temperature_ls,
+           temperature_data);
+  // Compute the actual temperature in Celsius
+  int actual_temp = 55 + ((temperature_data - 16384) / 160);
+  LOG_INFO("Temperature is: %u", actual_temp);
+  LOG_INFO("End of temperature test...");
+  if (-40 < actual_temp && actual_temp < 125)
+  {
+    result &= true;
```
kammce:
```suggestion
    result = true;
```
Like, why?

delwin-lei: I did this because I made the first result assignment during the device id check and I wanted to confirm that the values from the temperature sensor were valid as well. Assuming the temperature falls out of those bounds, we can say that the sensor is not working. If damage to the temperature sensor always impacts the ability to read from the device id register then I will replace the &=.

firmware/library/L5_Testing/factory_test.hpp
```diff
@@ -0,0 +1,229 @@
...
+      printf("Starting External Flash Test...\n\n");
+      Ssp ssp2(Ssp::Peripheral::kSsp2);
+      Gpio cs(1, 10);
+      bool result = false;
+      cs.SetAsOutput();
+      cs.SetHigh();
+
+      ssp2.SetPeripheralMode(Ssp::MasterSlaveMode::kMaster, Ssp::FrameMode::kSpi,
+                             Ssp::DataSize::kEight);
+      ssp2.SetClock(false, false, 1, 48);
```
kammce: Lets set this clock so to something much slower like:
```suggestion
      ssp2.SetClock(false, false, 100, 48);
```

firmware/library/L5_Testing/factory_test.hpp
```diff
@@ -0,0 +1,229 @@
...
+      ssp2.Initialize();
+      Delay(1000);
```
kammce:
```suggestion
```

kammce: This delay isn't really needed. Neither is the Delay(1) below, but keep it just to allow the cs pin to settle before you communicate.

firmware/library/L5_Testing/factory_test.hpp
```diff
@@ -0,0 +1,229 @@
...
+      // Read Manufacturer ID
+      ssp2.Transfer(0x9F);
+      array[0] = static_cast<uint8_t>(ssp2.Transfer(0x00));
+      array[1] = static_cast<uint8_t>(ssp2.Transfer(0x00));
+      array[2] = static_cast<uint8_t>(ssp2.Transfer(0x00));
+      array[3] = static_cast<uint8_t>(ssp2.Transfer(0x00));
+      LOG_INFO("Returned 0x%02X 0x%02X 0x%02X 0x%02X", array[0], array[1], array[2],
+               array[3]);
+      cs.SetHigh();
+      Delay(1000);
```
kammce:
```suggestion
      Delay(1);
```

kammce: Same as above. You only need 1ms for stuff to settle.

firmware/library/L5_Testing/factory_test.hpp
```diff
@@ -0,0 +1,229 @@
...
+
+    bool GestureSensorTest()
+    {
+      // Gesture Sensor Test
+      printf("++++++++++++++++++++++++++++++++++++++\n\n");
+      printf("Starting Gesture Sensor Test...\n\n");
+      constexpr uint8_t kApds9960IdRegisterAddress = 0x92;
+      bool result                                  = false;
+
+      uint8_t initialization_sequence[] = { 0x80, 0x05 };
```
kammce: Do we need to initialize the gesture sensor to get its ID?

firmware/library/L5_Testing/factory_test.hpp
```diff
@@ -0,0 +1,229 @@
...
+      printf("Starting Temperature Test...\n\n");
+      constexpr uint8_t kTemperatureIdAddress = 0xC0;
+      bool result                             = false;
+
+      result = CheckDeviceId(i2c, kTemperatureAddress, kTemperatureIdAddress, 0x14);
+
+      int actual_temp = CalculateTemperature(i2c);
+      LOG_INFO("Temperature is: %u", actual_temp);
+      LOG_INFO("End of temperature test...");
+      if (-40 < actual_temp && actual_temp < 125)
```
kammce:
```suggestion
      bool temp_test_result = (-40 < actual_temp && actual_temp < 125);
```

firmware/library/L5_Testing/factory_test.hpp
```diff
@@ -0,0 +1,229 @@
...
+      {
+        result = true;
+      }
+
+      return result;
```
kammce: return (id_test_result && temp_test_result);

Issue #303, opened 2019-01-07
Refactor SystemController's PowerOn Enumeration


kammce:
Currently a standard interface for SystemControllers cannot be made because of the PowerOnPeripheral() method. Without the interface, it takes an enumeration class defined within the specific implementation of the SystemController (for example Lpc40xxSystemController). To move that method to the interface would require that the SystemController know about the enum in the child class which is not entirely possible without having to flood the system controller with every single enum for every single implementation. Not only this, but enumeration classes are not polymorphic, meaning they cannot inherit from each other.

The solution to this is to move away from using enumerations as strong types that represent a peripheral and to use struct types, so they can be polymorphically inherited and passed to the PowerUpPeripheral() method.

```
struct SystemController
{
  struct PeripherialEnum
  {
    constexpr PeripherialEnum() {}
    uint32_t kDevice = -1;
  };

  template<size_t device_id>
  struct AddPeripheral : public PeripherialEnum
  {
    constexpr AddPeripheral() { kDevice = device_id; }
    constexpr operator PeripherialEnum()
    {
      return device_id;
    }
  };

  virtual void PowerUpPeripheral(PeripherialEnum enum_param) = 0;
};
```

PR #306, opened 2019-01-09
Refactor SystemController's PowerOn Enumeration

kammce:
Moved from enum class to a Peripheral struct type to make creating new
peripheral types easier as we expand to other platforms.


kammce:
Resolves #303 when pulled in.

PR #311, opened 2019-01-11
Adding bit utlity implementation with tests

kammce:
Tests also test that the bit:: utils can be used at compile time.


kammce:
This PR is stale and the contributor hasn't been responding. Closing this PR.

Issue #296, opened 2019-01-05
Add seperate make targets for user tests and full system tests


kammce:
Currently to test anything you must use `make test`, which will pull in all of the source files from the library folder, makes compiling tests extremely long and makes running test longer, although its runtime effect is very negligible. 

I propose a `make library-test` which pulls in the library files and `make test` simply pulls in the test source files from the included `project.mk` file.

kammce:
Resolved once #314 is merged.

PR #314, opened 2019-01-12
Adding "make user-test"

kammce:
Now you can test a select number of test files, rather than having to
compile all of the test files, you can select use a project.mk file
within your projects folder to include only the set of test files you
want to comple.


kammce:
Resolves #296 once this is merged.

PR #319, opened 2019-01-13
Acceleration Sensor Implementation

diarkia124:
The use of the I2C class functions are there, however, there is nothing that comes out of the SDA or SCL pins when using the Accelerometer Class unless under some special conditions.


diarkia124:
Yeah maybe a new branch was a better idea :/

PR #331, opened 2019-01-18
Add New Temperature Sensor interface.

ChuefengVang:
Removed method "GetTemperatureBytes()" as it is not needed anymore for new temperature sensor device.

### DIFF
--- a/firmware/library/L2_HAL/sensors/temperature/temperature.hpp
+++ b/firmware/library/L2_HAL/sensors/temperature/temperature.hpp
@@ -1,11 +1,12 @@
-// Interface used to sense the temperature of the temperature sensor.
 #pragma once
-#include <cstdint>
+
+#include "L1_Drivers/i2c.hpp"
+
 class TemperatureInterface
 {
  public:
-  virtual bool Initialize() = 0;
-  virtual uint32_t GetTemperatureBytes() = 0;
-  virtual float GetCelsius() = 0;
+  virtual bool Initialize()     = 0;
+  virtual float GetCelsius()    = 0;
   virtual float GetFahrenheit() = 0;
 };
+


kammce:
Split this into two PRs. One for just the interface. And the other for the LPC40xx implementation.

ChuefengVang:
Done.

firmware/library/L2_HAL/temperature.hpp
```diff
@@ -0,0 +1,14 @@
...
+
+#include <cstdint>
+#include <cstdio>
+#include "L1_Drivers/i2c.hpp"
+
+class TemperatureInterface
+{
+ public:
+  virtual bool Initialize()                       = 0;
+  virtual int GetCelsius()                        = 0;
```
kammce: What size int are you using for this? Please use a specific integer size for the values here. Please see <cstdint> for more details.

firmware/library/L2_HAL/temperature.hpp
```diff
@@ -0,0 +1,14 @@
+#pragma once
+
+#include <cstdint>
+#include <cstdio>
+#include "L1_Drivers/i2c.hpp"
```
kammce: You aren't using any of these includes. Including them will only slow down compilation time. Do not include what are not using. Always include what you are using.

With my comment below, you should keep <cstdint> though.

firmware/library/L2_HAL/temperature.hpp
```diff
@@ -0,0 +1,14 @@
...
+
+class TemperatureInterface
+{
+ public:
+  virtual bool Initialize()                       = 0;
+  virtual int GetCelsius()                        = 0;
+  virtual int GetFahrenheit(int temperature_data) = 0;
```
kammce:
```suggestion
  virtual int GetFahrenheit() = 0;
```

ChuefengVang: Did all the changes and moved the file over to L2_HAL/sensors/temperature.  Also removed the example main.cpp but couldn't remove the folder. Should be fine there I'll include the example after implementation is pull.

PR #335, opened 2019-01-19
Add Temperature sensor implementation.

ChuefengVang:
The repo's name should be Implementation but it is fine.


firmware/library/L2_HAL/temperature.hpp
```diff
@@ -0,0 +1,64 @@
...
+  virtual int GetFahrenheit(int temperature_data) = 0;
+};
+
+class Temperature : public TemperatureInterface
+{
+ public:
+  bool Initialize() override
+  {
+  temperature_device_.Initialize();
+  uint8_t device_id = 0x00;
```
kammce:
```suggestion
  uint8_t temperature_sensor_id = 0x00;
```

firmware/library/L2_HAL/temperature.hpp
```diff
@@ -0,0 +1,64 @@
...
+  // Checks to see if the value read by I2c
+  // for the Temperature Chip ID register
+  // is the same as the Base ID (0x14).
+  temperature_device_.WriteThenRead(kTemperatureAddress, &kTemperatureId,
+  1, &device_id, 1);
+  return(device_id == kMemoryMapAddress);
```
kammce:
```suggestion
  return (device_id == kMemoryMapAddress);
```

firmware/library/L2_HAL/temperature.hpp
```diff
@@ -0,0 +1,64 @@
...
+  }
+
+  int GetCelsius() override
+  {
+  uint8_t const kOneBurst [2]   = { 0xC4, 0x04 };
```
kammce:
```suggestion
  const uint8_t kOneBurstRegister = 0xC4;
```

firmware/library/L2_HAL/temperature.hpp
```diff
@@ -0,0 +1,64 @@
...
+  uint8_t const kAutoincBit [2] = { 0xC5, 0x01 };
+  // The register will enable the device to collect data once
+  // and automatically sets the stop bit to 0 (2nd bit).
+  status_ = temperature_device_.Write(kTemperatureAddress, kOneBurst, 2);
```
kammce:
```suggestion
  status_ = temperature_device_.Write(kTemperatureAddress, { kOneBurst 0x04 }, 2);
```

firmware/library/L2_HAL/temperature.hpp
```diff
@@ -0,0 +1,64 @@
...
+  }
+
+  int GetFahrenheit(int temperature_data)
+  {
+  // Formula used to convert Celsius to Fahrenheit
+  return (((temperature_data - 32) * 5) / 9 );
+  }
+
+ private:
+  I2c temperature_device_;
```
kammce: A better name would be `i2c_`

firmware/library/L2_HAL/temperature.hpp
```diff
@@ -0,0 +1,64 @@
...
+  Status status_;
+  static constexpr uint8_t kTemperatureAddress = 0x31;
+  static constexpr uint8_t kTemperatureId      = 0xC0;
+  static constexpr uint8_t kMemoryMapAddress   = 0x14;
```
kammce: static and constexpr should be at the top.

https://google.github.io/styleguide/cppguide.html#Class_Format

firmware/library/L2_HAL/temperature.hpp
```diff
@@ -0,0 +1,64 @@
...
+  {
+  uint8_t const kOneBurst [2]   = { 0xC4, 0x04 };
+  uint8_t const kAutoincBit [2] = { 0xC5, 0x01 };
+  // The register will enable the device to collect data once
+  // and automatically sets the stop bit to 0 (2nd bit).
+  status_ = temperature_device_.Write(kTemperatureAddress, kOneBurst, 2);
+  // Auto increments I2c register address pointer.
+  status_ = temperature_device_.Write(kTemperatureAddress, kAutoincBit, 2);
+  uint8_t least_significant_register = 0xC2;
+  uint8_t most_significant_register  = 0xC1;
```
kammce: Make these into constexprs at the top of the method.

firmware/library/L2_HAL/temperature.hpp
```diff
@@ -0,0 +1,64 @@
...
+  temperature_device_.WriteThenRead(kTemperatureAddress,
+ &most_significant_register, 1, &most_significant_register, 1);
+  temperature_device_.Read(kTemperatureAddress, &least_significant_register, 1);
+
+  int data = 0;
+  data = ((most_significant_register & 0x7F) << 8) | least_significant_register;
+  // The required computation after bit shifting.
+  return (((data - 16384) / 160 ) + 55);
```
kammce: Put a comment here referencing where this equation came from.

kammce:
```suggestion
  return (((data - (1 << 14)) / 160 ) + 55);
```

firmware/library/L2_HAL/temperature.hpp
```diff
@@ -0,0 +1,64 @@
...
+  // Auto increments I2c register address pointer.
+  status_ = temperature_device_.Write(kTemperatureAddress, kAutoincBit, 2);
+  uint8_t least_significant_register = 0xC2;
+  uint8_t most_significant_register  = 0xC1;
+  temperature_device_.WriteThenRead(kTemperatureAddress,
+ &most_significant_register, 1, &most_significant_register, 1);
+  temperature_device_.Read(kTemperatureAddress, &least_significant_register, 1);
+
+  int data = 0;
+  data = ((most_significant_register & 0x7F) << 8) | least_significant_register;
```
kammce: Prefer to use the `utility/bit.hpp` calls for this sort of bit manipulation.

firmware/library/L2_HAL/temperature.hpp
ChuefengVang: Update on this portion. I left these as is because I2c did not like them as constexprs. That would make sense since the values are being over written.

firmware/library/L2_HAL/sensors/temperature/temperature.hpp
```diff
@@ -1,11 +1,74 @@
...
+  return (fahrenheit_data_);
+  }
+
+ private:
+  static constexpr uint8_t kTemperatureAddress = 0x31;
+  static constexpr uint8_t kTemperatureId      = 0xC0;
+  static constexpr uint8_t kMemoryMapAddress   = 0x14;
+  float celsius_data_;
+  float fahrenheit_data_;
+  I2c i2c_;
```
kammce: See the `button.hpp` class to get an idea of how to use L1_Driver layer objects.

firmware/library/L2_HAL/sensors/temperature/temperature.hpp
```diff
@@ -1,11 +1,74 @@
...
+  // Formula used to convert Celsius to Fahrenheit
+  // Note that: 1 << 5 = 2^5 = 32
+  fahrenheit_data_ = ((celsius_data_ - (1 << 5)) * 5) / 9;
+  return (fahrenheit_data_);
+  }
+
+ private:
+  static constexpr uint8_t kTemperatureAddress = 0x31;
+  static constexpr uint8_t kTemperatureId      = 0xC0;
+  static constexpr uint8_t kMemoryMapAddress   = 0x14;
```
kammce: These constexpr values should be at the top just like the other classes have then and within the public section. No need to hide them in the private section.

firmware/library/L2_HAL/sensors/temperature/temperature.hpp
```diff
@@ -1,11 +1,74 @@
...
+  celsius_data_ = ((bit_data - (1 << 14)) / 160) + 55;
+  return (celsius_data_);
+  }
+
+  float GetFahrenheit() override
+  {
+  // Formula used to convert Celsius to Fahrenheit
+  // Note that: 1 << 5 = 2^5 = 32
+  fahrenheit_data_ = ((celsius_data_ - (1 << 5)) * 5) / 9;
+  return (fahrenheit_data_);
```
kammce: See this section of the google coding guidelines:
https://google.github.io/styleguide/cppguide.html#Return_Values 

```suggestion
  return ((GetCelsius() - (1 << 5)) * 5) / 9;
```

firmware/library/L2_HAL/sensors/temperature/temperature.hpp
```diff
@@ -1,11 +1,74 @@
...
+  }
+
+ private:
+  static constexpr uint8_t kTemperatureAddress = 0x31;
+  static constexpr uint8_t kTemperatureId      = 0xC0;
+  static constexpr uint8_t kMemoryMapAddress   = 0x14;
+  float celsius_data_;
+  float fahrenheit_data_;
```
kammce: These floats should not exist in the object. If you look at how you are using them, you can see that you are only use them as storage containers for the duration of the method. Fahrenheit is only used within the GetFarenheit() method, but no where else. Which means that every instance of this object takes up 8 additional bytes of data that isn't really be used for anything. 

Another thing to keep in mind is that you shouldn't cache sensor data. When the user runs GetCelsius(), they are expect to get the current sensor information. If you give them a cached version of the data, they will not be happy with your driver. 

With the size cost and caching issue in mind, keeping these variables in this class unnecessary increases its size without any added benefit.

firmware/library/L2_HAL/sensors/temperature/temperature.hpp
```diff
@@ -1,11 +1,74 @@
...
+  I2c i2c_;
+  Status status_;
```
kammce: You never give the user the option to retrieve the status information nor do you actually use it in your code. You assign it, but never read it. So keeping this within the class doesn't do anything but increase the size of the object. Bare in mind that ever single byte on the MCU counts. We cannot afford unnecessary adding weight to each object when we only have 96kB of ram total.

firmware/library/L2_HAL/sensors/temperature/temperature.hpp
kammce: Also, why cache the value when you can just get it by calling the GetCelsius() method instead.

firmware/library/L2_HAL/sensors/temperature/temperature.hpp
```diff
@@ -1,11 +1,74 @@
...
+  1, &temperature_sensor_id, 1);
+  return (temperature_sensor_id == kMemoryMapAddress);
+  }
+
+  float GetCelsius(I2c i2c_temperature) override
+  {
+  uint8_t one_burst_register         = 0xC4;
+  uint8_t autoinc_bit_register       = 0xC5;
+  uint8_t most_significant_register  = 0xC1;
+  uint8_t least_significant_register = 0xC2;
```
kammce: Please make these all constexpr.

firmware/library/L2_HAL/sensors/temperature/temperature.hpp
```diff
@@ -1,11 +1,74 @@
...
+ { one_burst_register, 0x04 }, 2);
+  // Auto increments I2c register address pointer.
+  status_ = i2c_temperature.Write(kTemperatureAddress,
+ { autoinc_bit_register, 0x01 }, 2);
+  i2c_temperature.WriteThenRead(kTemperatureAddress,
+ &most_significant_register, 1, &most_significant_register, 1);
+  i2c_temperature.Read(kTemperatureAddress, &least_significant_register, 1);
+
+  int bit_data = 0;
+  bit_data = ((most_significant_register & 0x7F) << 8);
```
kammce: Why not use `bit::Insert` here as well? The point is to remove the need for you to do the bit manipulation.

firmware/library/L2_HAL/sensors/temperature/temperature.hpp
```diff
@@ -1,11 +1,74 @@
...
+  // Note that: 1 << 14 = 2^14 = 16384
+  celsius_data_ = ((bit_data - (1 << 14)) / 160) + 55;
+  return (celsius_data_);
+  }
+
+  float GetFahrenheit() override
+  {
+  // Formula used to convert Celsius to Fahrenheit
+  // Note that: 1 << 5 = 2^5 = 32
+  fahrenheit_data_ = ((celsius_data_ - (1 << 5)) * 5) / 9;
```
kammce:
```suggestion
  fahrenheit_data_ = ((celsius_data_ - 32) * 5) / 9;
```
So just because "32" is a power of 2, doesn't mean that you should always transform them into shifts. The reason for the value above required context. 1 << 14 is the way it is due to how the bits are setup in the temperature register. In this case, it is know that you must subtract 32 within the calculation to get Fahrenheit. 

So looking at each case you can see that, 16384 is less obvious then 1 << 14 with respect to bit manipulation, and that (1 << 5) is less obvious then 32 when you are dealing with the conversion from Celcius to Fahrenheit.

firmware/library/L2_HAL/sensors/temperature/temperature.hpp
```diff
@@ -1,11 +1,74 @@
...
+  // Auto increments I2c register address pointer.
+  status_ = i2c_temperature.Write(kTemperatureAddress,
+ { autoinc_bit_register, 0x01 }, 2);
+  i2c_temperature.WriteThenRead(kTemperatureAddress,
+ &most_significant_register, 1, &most_significant_register, 1);
+  i2c_temperature.Read(kTemperatureAddress, &least_significant_register, 1);
+
+  int bit_data = 0;
+  bit_data = ((most_significant_register & 0x7F) << 8);
+  bit_data = bit::Insert(bit_data, least_significant_register, 0, 128);
```
kammce: C++ does not support integers of width 128. Re-read the comment documation in the bit.hpp file for bit::Insert.

firmware/library/L2_HAL/sensors/temperature/temperature.hpp
```diff
@@ -1,11 +1,74 @@
...
+  status_ = i2c_temperature.Write(kTemperatureAddress,
+ { one_burst_register, 0x04 }, 2);
+  // Auto increments I2c register address pointer.
+  status_ = i2c_temperature.Write(kTemperatureAddress,
+ { autoinc_bit_register, 0x01 }, 2);
+  i2c_temperature.WriteThenRead(kTemperatureAddress,
+ &most_significant_register, 1, &most_significant_register, 1);
+  i2c_temperature.Read(kTemperatureAddress, &least_significant_register, 1);
+
+  int bit_data = 0;
```
kammce:
```suggestion
  uint16_t bit_data = 0;
```
Please use stdint rather then using int. Please see the style guidelines for SJSU-Dev2 here:
https://sjsu-dev2.readthedocs.io/en/latest/rst/contributing-toc.html

Find the section regarding the proper declaration of integers.

firmware/library/L2_HAL/sensors/temperature/temperature.hpp
ChuefengVang: Just read your bottom comment. That does make sense. I think I'll try to approach it with the way I had before then.

firmware/library/L2_HAL/sensors/temperature/temperature.hpp
ChuefengVang: I think this portion is a bit complicated. I'll go through with you on this.

firmware/library/L2_HAL/sensors/temperature/temperature.hpp
kammce:
```
bit_data = bit::Insert(bit_data, lsb, 0, 8);
bit_data = bit::Insert(bit_data, msb, 8, 8);
bit_data = bit::Clear(bit_data, 15);
```

ChuefengVang:
See up to date pull request: https://github.com/kammce/SJSU-Dev2/pull/339

PR #361, opened 2019-01-26
Adding Periodic Scheduler

HuangDave:
- Added PeriodicScheduler class to schedule  tasks to run at 1Hz, 10Hz, 100Hz, and 1000Hz,
- Increased the maximum priority levels for FreeRTOS from 5 to 9 to add 4 more priority
  levels for the periodic tasks.
- Added the following fake functions to rtos.hpp for unit testing:
	- xQueueGenericCreateStatic
	- xQueueGenericSend
	- xQueueSemaphoreTake

### DIFF
--- a/firmware/library/L3_Application/task.hpp
+++ b/firmware/library/L3_Application/task.hpp
@@ -132,6 +132,7 @@ class Task : public TaskInterface
   }
   bool DeclaredOnStackCheck()
   {
+#if !defined(HOST_TEST)
     // This task's position in memory
     intptr_t address = reinterpret_cast<intptr_t>(this);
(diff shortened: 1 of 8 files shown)


firmware/library/L3_Application/task.hpp
```diff
@@ -39,7 +39,7 @@ class TaskInterface
 };
 
 template <size_t kStackSize>
-class Task : public TaskInterface
+class Task : public virtual TaskInterface
```
kammce: Wait, why is this needed? Virtual inheritance can be a very nasty thing to use.

kammce:
Also, this looks really good. Can't wait to pull it in.

PR #362, opened 2019-01-26
Functioning timer driver with 4 timers and interrupt example.

RadConMannBor:
Configured to work with interrupts and accurate to the millisecond.

### DIFF
--- a/firmware/examples/Timer/source/main.cpp
+++ b/firmware/examples/Timer/source/main.cpp
@@ -1,0 +1,60 @@
+// Hardware Timer Driver example that provides example of
+// how to use the hardware timer driver.
+#include "L0_LowLevel/interrupt.hpp"
+#include "L0_LowLevel/LPC40xx.h"
+#include "L1_Drivers/timer.hpp"
+#include "utility/log.hpp"
+#include "utility/time.hpp"
+
+Timer timer0(TimerInterface::kTimer0);
+Timer timer1(TimerInterface::kTimer1);
+Timer timer2(TimerInterface::kTimer2);
+Timer timer3(TimerInterface::kTimer3);
+
+// Put whatever functions you want to be triggered within these ISR's
+// and you can use the timer to trigger whatever event you want.
+void Timer0ISR(void)
+{
+  printf("Successfully triggered interrupt from timer 0\n\n");
+  timer0.ClearInterrupts();
+}
+void Timer1ISR(void)
+{
+  printf("Successfully triggered interrupt from timer 1.\n\n");
+  timer1.ClearInterrupts();
+}
+void Timer2ISR(void)
+{
+  printf("Successfully triggered interrupt from timer 2\n\n");
+  timer2.ClearInterrupts();
+}
+void Timer3ISR(void)
+{
+  printf("Successfully triggered interrupt from timer 3\n\n");
+  timer3.ClearInterrupts();
+}
+
+void StartDemo(Timer * timer_demo0, Timer * timer_demo1, Timer * timer_demo2,
+               Timer * timer_demo3)
+{
+  timer_demo0->SetTimer(1000000, Timer::kMat0, Timer::kInterruptRestart);
+  timer_demo1->SetTimer(2000000, Timer::kMat1, Timer::kInterruptRestart);
+  timer_demo2->SetTimer(3000000, Timer::kMat2, Timer::kInterruptRestart);
+  timer_demo3->SetTimer(4000000, Timer::kMat3, Timer::kInterruptRestart);
+}
+
+// In this example all timers are initilialized to 1 micro second per tick
+int main(void)
+{
+  timer0.Initialize(Timer::TimerPort::kTimer0, 1, Timer0ISR, 1);
+  timer1.Initialize(Timer::TimerPort::kTimer1, 1, Timer1ISR, 1);
+  timer2.Initialize(Timer::TimerPort::kTimer2, 1, Timer2ISR, 1);
+  timer3.Initialize(Timer::TimerPort::kTimer3, 1, Timer3ISR, 1);
+
+  LOG_INFO("\n");
+  LOG_INFO("Start Timer Demo\n");
+  while (true)
+  {
+    StartDemo(&timer0, &timer1, &timer2, &timer3);
+  }
+}
--- a/firmware/library/L1_Drivers/timer.hpp
+++ b/firmware/library/L1_Drivers/timer.hpp
@@ -1,0 +1,133 @@
+#pragma once
+#include <cstdint>
+#include <limits>
+
+#include "L0_LowLevel/interrupt.hpp"
+#include "L0_LowLevel/LPC40xx.h"
+#include "L0_LowLevel/system_controller.hpp"
+#include "L1_Drivers/pin.hpp"
+#include "utility/enum.hpp"
+#include "utility/log.hpp"
+
+class TimerInterface
+{
+ public:
+  enum RegisterChoice : uint8_t
+  {
+    kRegMR0 = 0,
+    kRegMR1 = 1,
+    kRegMR2 = 2,
+    kRegMR3 = 3,
+    kRegCR0 = 4,
+    kRegCR1 = 5
+  };
+  enum TimerPort : uint8_t
+  {
+    kTimer0 = 0,
+    kTimer1 = 1,
+    kTimer2 = 2,
+    kTimer3 = 3
+  };
+  enum MatchControlRegister : uint8_t
+  {
+    kMat0 = 0,
+    kMat1 = 3,
+    kMat2 = 6,
+    kMat3 = 9
+  };
+  // Modes for each match register that when match register matches timer
+  // system can trigger an Interrupt, Stop Counter, or Restart and in any
+  // mixture of those commands.
+  enum TimerSetting : uint8_t
+  {
+    kInterrupt            = 0b001,
+    kRestart              = 0b010,
+    kStop                 = 0b100,
+    kInterruptRestart     = 0b011,
+    kInterruptStop        = 0b101,
+    kRestartStop          = 0b110,
+    kInterruptRestartStop = 0b111
+  };
+  virtual void Initialize(TimerPort timer, uint32_t us_per_tick, IsrPointer isr,
+                          int32_t priority) = 0;
+  virtual void SetTimer(uint32_t time, MatchControlRegister reg,
+                        TimerSetting mode)  = 0;
+  virtual uint32_t GetTimer()               = 0;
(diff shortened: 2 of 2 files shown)


firmware/library/L1_Drivers/timer.hpp
```diff
@@ -0,0 +1,157 @@
...
+    mat1_->SetAsAnalogMode(true);
+    mat1_->SetMode(PinInterface::Mode::kInactive);
+    mat2_->SetPinFunction(kTime0Mat0);
+    mat2_->SetAsAnalogMode(true);
+    mat2_->SetMode(PinInterface::Mode::kInactive);
+    mat3_->SetPinFunction(kTime0Mat0);
+    mat3_->SetAsAnalogMode(true);
+    mat3_->SetMode(PinInterface::Mode::kInactive);
+    // set Prescale register for Prescale Counter to milliseconds
+    tim_register[static_cast<uint8_t>(timer)]->PR &= ~(kClear << 1);
```
kammce: Be sure to include `util/enum.hpp`
```suggestion
   uint8_t timer_index = util::Value(timer);
    tim_register[]->PR &= ~(kClear << 1);
```

kammce: Do this for all other instances of static_cast of enum classes.

firmware/library/L1_Drivers/timer.hpp
```diff
@@ -0,0 +1,157 @@
...
+    tim_register[static_cast<uint8_t>(timer)]->TCR |= (1 << 0);
+  }
+
+  // Function that sets a timer of your choice (0-3) to a specific
+  // time in milliseconds,and designates the mode of
+  // functionality you want from it, interrupt, stop, or reset.
+  // --NOTICE-- Currently Each Timer is directly hooked to match
+  void SetTimer(uint32_t time, RegisterChoice reg, TimerSetting mode) override
+  {
+        switch (reg)
```
kammce: NIT: fix the formatting. Use `clang-format` on these files.

firmware/library/L1_Drivers/timer.hpp
```diff
@@ -0,0 +1,157 @@
...
+    {
+        case kRegMR0:
```
kammce: Use a lookup table rather then copying the same lines of code 4 times. The are all the same except that you shift by kMatX.

firmware/library/L1_Drivers/timer.hpp
```diff
@@ -0,0 +1,157 @@
...
+            tim_register[channel_]->MR3 |= (((time/2)*1000) << 0);
+            break;
+        default:
+            break;
+    }
+  }
+
+  uint32_t GetTimer() override
+  {
+    return(tim_register[channel_]->TC);
```
kammce:
```suggestion
    return tim_register[channel_]->TC;
```
Never EVER make return look like a function call.
https://google.github.io/styleguide/cppguide.html#Return_Values

firmware/library/L1_Drivers/timer.hpp
```diff
@@ -0,0 +1,157 @@
...
+  }
+
+  void ClearInterrupts() override
+  {
+    tim_register[channel_]->IR |= (1 << kRegMR0);
+    tim_register[channel_]->IR |= (1 << kRegMR1);
+    tim_register[channel_]->IR |= (1 << kRegMR2);
+    tim_register[channel_]->IR |= (1 << kRegMR3);
```
kammce: Should you be clearing every interrupt in this case? Or does it not matter, because clearing an interrupt that hasn't fired doesn't prevent it from firing a clock later?

firmware/library/L1_Drivers/timer.hpp
```diff
@@ -0,0 +1,157 @@
...
+            break;
+    }
+  }
+
+  uint32_t GetTimer() override
+  {
+    return(tim_register[channel_]->TC);
+  }
+
+  void ClearInterrupts() override
```
kammce: Make all overrides final
```suggestion
  void ClearInterrupts() override final
```

firmware/library/L1_Drivers/timer.hpp
```diff
@@ -0,0 +1,157 @@
...
+            tim_register[channel_]->MCR &= ~(kClearMode << kMat3);
+            tim_register[channel_]->MCR |= (mode << kMat3);
+            tim_register[channel_]->MR3 |= (((time/2)*1000) << 0);
+            break;
+        default:
+            break;
+    }
+  }
+
+  uint32_t GetTimer() override
```
kammce:
```suggestion
  [[gnu::always_inline]] uint32_t GetTimer() override
```
This will optimize this function to force it to become unlined at the call site, rather than doing an actual function call.

firmware/library/L1_Drivers/timer.hpp
```diff
@@ -0,0 +1,157 @@
...
+    mat0_->SetMode(PinInterface::Mode::kInactive);
+    mat1_->SetPinFunction(kTime0Mat0);
+    mat1_->SetAsAnalogMode(true);
+    mat1_->SetMode(PinInterface::Mode::kInactive);
+    mat2_->SetPinFunction(kTime0Mat0);
+    mat2_->SetAsAnalogMode(true);
+    mat2_->SetMode(PinInterface::Mode::kInactive);
+    mat3_->SetPinFunction(kTime0Mat0);
+    mat3_->SetAsAnalogMode(true);
+    mat3_->SetMode(PinInterface::Mode::kInactive);
```
kammce: What if you do not want to use the match registers though? I think in most cases, most users do not want to use the match registers. the should be another method for turning these on rather than the main Init.

firmware/examples/Timer/source/main.cpp
```diff
@@ -0,0 +1,65 @@
+// Hardware Timer Driver example that provides example of
+// how to use the Hardware timer driver.
+#include "L0_LowLevel/interrupt.hpp"
+#include "L0_LowLevel/LPC40xx.h"
+#include "L1_Drivers/timer.hpp"
+#include "utility/log.hpp"
+#include "utility/time.hpp"
+// Input Number for starting demo
```
kammce: Delete.

firmware/examples/Timer/source/main.cpp
```diff
@@ -0,0 +1,65 @@
+// Hardware Timer Driver example that provides example of
+// how to use the Hardware timer driver.
```
kammce:
```suggestion
// how to use the hardware timer driver.
```

firmware/examples/Timer/source/main.cpp
```diff
@@ -0,0 +1,65 @@
...
+  RegisterIsr(TIMER0_IRQn, Timer0ISR, true, 1);
+  RegisterIsr(TIMER1_IRQn, Timer1ISR, true, 1);
+  RegisterIsr(TIMER2_IRQn, Timer2ISR, true, 1);
+  RegisterIsr(TIMER3_IRQn, Timer3ISR, true, 1);
+
+  timer0.Initialize(Timer::TimerPort::kTimer0);
+  timer1.Initialize(Timer::TimerPort::kTimer1);
+  timer2.Initialize(Timer::TimerPort::kTimer2);
+  timer3.Initialize(Timer::TimerPort::kTimer3);
+  DEBUG_PRINT("\n");
```
kammce: Do not use DEBUG_PRINT, use LOG_INFO.

firmware/library/L1_Drivers/timer.hpp
```diff
@@ -0,0 +1,157 @@
...
+  virtual uint32_t GetTimer() = 0;
+  virtual void ClearInterrupts() = 0;
+};
+
+class Timer : public TimerInterface, protected Lpc40xxSystemController
+{
+ public:
+  static constexpr uint8_t kTime0Mat0   = 0b011;
+  static constexpr uint8_t kClearMode   = 0b0111;
+  static constexpr uint8_t kPrescaleMicro   = 48;
```
kammce: What if the user chooses to change the system frequency to something else? This makes the assumption that the cpu speed is 48 MHz which may not be the case. Use `GetPeripheralFrequency()` in place of where you use the PreScalar. Also allow the user to select a prescaler rather than defining it for them. I would like to be able to control the prescalar's value. Let the developer figure out how to make a 1ms timer rather than trying to figure that out here.

firmware/library/L1_Drivers/timer.hpp
RadConMannBor: In the case of the IR register, sending a logic high to any of the registers will only reset that register and allow a new interrupt to be fired.  I have not experienced any cases where me sending a high will cause an false interrupt flag.

firmware/library/L1_Drivers/timer.hpp
```diff
@@ -0,0 +1,157 @@
...
+  static constexpr uint32_t kClear = 0b11111111111111111111111111111111;
```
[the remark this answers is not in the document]
RadConMannBor: What should I include for this command?  you just said i "need to include ."

firmware/library/L1_Drivers/timer.hpp
RadConMannBor: So i am trying to use a lookup table for the pointers to the match registers and I have tried to do it like this
`static volatile uint32_t * match[6] = 
               { &tim_register[channel_]->MR0, &tim_register[channel_]->MR1,
                 &tim_register[channel_]->MR2, &tim_register[channel_]->MR3,
                 &tim_register[channel_]->CR0, &tim_register[channel_]->CR1
               };`

However It says that channel_ is not declared in this scope.  I am sorry to ask you, but I am confused as to how I can create this lookup table or where I have made a mistake.

firmware/examples/Timer/source/main.cpp
```diff
@@ -0,0 +1,65 @@
...
+
+// In this example all timers are initilialized to 1 micro second per tick.
+int main(void)
+{
+  RegisterIsr(TIMER0_IRQn, Timer0ISR, true, 1);
+  RegisterIsr(TIMER1_IRQn, Timer1ISR, true, 1);
+  RegisterIsr(TIMER2_IRQn, Timer2ISR, true, 1);
+  RegisterIsr(TIMER3_IRQn, Timer3ISR, true, 1);
+
+  timer0.Initialize(Timer::TimerPort::kTimer0, 1);
```
kammce: Why doesn't Initialize handle `RegisterIsr` this for us?

RadConMannBor: I added it to Initialize.

RadConMannBor:
For some reason Codacy/PR Quality review is stalled and does not finish

Issue #180, opened 2018-09-05
SSP get/set registers


diarmadhi:
Change the current use of class variables to getter/setter functions. See pull request #164

PR #363, opened 2019-01-27
Ssp Updated Driver

diarmadhi:
This updated driver addresses the get/set registers from #180 as well as update/add the example on how to use the driver and unit test.

Unable to incorporate bit fields.

### DIFF
--- a/firmware/library/L1_Drivers/ssp.hpp
+++ b/firmware/library/L1_Drivers/ssp.hpp
@@ -13,7 +13,7 @@
 /// Note that all register modifications must be made before the SSP
 /// is enabled in the CR1 register (see page 612 of user manual UM10562)
 /// If changes are desired after the Initialize function is called, the
-/// peripheral must be disable, and then re-enabled after changes are made.
+/// peripheral must be disabled, and then re-enabled after changes are made.
 
 #pragma once
 
@@ -61,6 +61,8 @@ class SspInterface
   virtual void Initialize()                     = 0;
   virtual bool GetTransferStatus()              = 0;
   virtual uint16_t Transfer(uint16_t data)      = 0;
+  // virtual void TxData(uint16_t data)            = 0;
+  // virtual uint16_t RxData(uint16_t data)        = 0;
   virtual void SetSpiMasterDefault()            = 0;
   virtual void SetPeripheralMode(MasterSlaveMode mode, FrameMode frame,
                                  DataSize size) = 0;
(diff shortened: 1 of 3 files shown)


firmware/library/L1_Drivers/ssp.hpp
```diff
@@ -246,27 +213,113 @@ class Ssp final : public SspInterface, protected Lpc40xxSystemController
...
   }
-  // Sets the various modes for the Peripheral
-  // @param mode - master or slave mode
-  // @param frame - format for Peripheral data to use
-  // @param size - number of bits per frame
-  void SetPeripheralMode(MasterSlaveMode mode, FrameMode frame,
-                         DataSize size) override
+
+  /// Sets up SSP peripheral as SPI master
+  void SetSpiMasterDefault() override
```
kammce: SetSpiMasterDefault shouldn't be doing any bit manipulation. As you can see, you are doing double duty here and writing the same set of code twice. You can simply call/reuse the methods you have already made rather than doing the direct register manipluation.

firmware/library/L1_Drivers/test/ssp_test.cpp
```diff
@@ -1,12 +1,83 @@
...
+  // Set up Mock for PinCongiure
+  Mock<PinInterface> mock_mosi;
+  Mock<PinInterface> mock_miso;
+  Mock<PinInterface> mock_sck;
+
+  Fake(Method(mock_mosi, SetPinFunction));
+  Fake(Method(mock_miso, SetPinFunction));
+  Fake(Method(mock_sck, SetPinFunction));
+
+
```
kammce: Remove extra line.

firmware/library/L1_Drivers/ssp.hpp
```diff
@@ -246,27 +213,113 @@ class Ssp final : public SspInterface, protected Lpc40xxSystemController
...
   }
 
-  // Sets the clock rate for the Peripheral
-  void SetClock(bool polarity, bool phase, uint8_t divider,
-                uint8_t prescaler) override
+  /// Sets the various modes for the Peripheral
+  /// @param mode - master or slave mode
+  /// @param frame - format for Peripheral data to use
+  /// @param size - number of bits per frame
+  void SetPeripheralMode(MasterSlaveMode mode,
```
kammce: For this method and the rest that do bit manipulation, I believe that the best way forward for this control register is to create a bitfield overlay. See `L1/can.hpp` and how a bit field is used with the register.

firmware/library/L1_Drivers/ssp.hpp
```diff
@@ -184,57 +176,32 @@ class Ssp final : public SspInterface, protected Lpc40xxSystemController
...
-    ssp_registers[pssp]->CR1 |= (master_mode_ << kMasterModeBit);
   }
 
-  // Checks if the SSP controller is idle.
-  // @return 1 - the controller is sending or receiving a data frame.
-  // @return 0 - the controller is idle.
+  /// Checks if the SSP controller is idle.
+  /// @return 1 - the controller is sending or receiving a data frame.
+  /// @return 0 - the controller is idle.
   bool GetTransferStatus() override
```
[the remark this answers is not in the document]
diarmadhi: `IsBusy()` sounds ambiguous to me. I mean what are you checking to see if it is busy? Would `IsTransferRegBusy()` be acceptable?

firmware/library/L1_Drivers/ssp.hpp
diarmadhi: Oh, like the register mapping from CMPE146? I can only work on this in the evenings, what's our dead date?

firmware/library/L1_Drivers/ssp.hpp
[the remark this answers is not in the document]
kammce: I feel that that is confusing for a beginning in embedded. I think `IsTransferBusy()` would be a good compromise as it abstracts from the fact that you are reading from a register.

firmware/library/L1_Drivers/ssp.hpp
kammce: Yeah, same here. The due date was last weekend, but the absolute final time is this weekend since I will need to lock the repo as the semester starts. I don't want to require that students reinstall or pull from git in CmpE 30.

PR #379, opened 2019-02-06
Adding cp2102 driver to instructions

kammce:


kammce:
Commit was merged in via PR #380

PR #481, opened 2019-05-27
Changing PeriodicScheduler to use FreeRTOS software timers

HuangDave:
- Added vApplicationGetTimerTaskMemory() to startup.cpp.
- Configured FreeRTOSConfig.h to enable the use of software timers.
- Updated Unit Tests for PeriodicScheduler
   - Added fake functions for xTimerCreateStatic, xTimerGenericCommand,
     and vTimerGetTimerID, vTimerSetTimerID.

### DIFF
--- a/library/L3_Application/test/periodic_scheduler_test.cpp
+++ b/library/L3_Application/test/periodic_scheduler_test.cpp
@@ -21,11 +21,8 @@ namespace sjsu
 {
 namespace rtos
 {
-
 TEST_CASE("Testing PeriodicTask", "[periodic_task]")
 {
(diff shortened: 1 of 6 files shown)


library/L3_Application/test/periodic_scheduler_test.cpp
```diff
@@ -53,6 +52,10 @@ TEST_CASE("Testing PeriodicScheduler", "[periodic_scheduler]")
 {
   using namespace rtos;  // NOLINT
```
kammce: NIT: Using namespace is no longer necessary since the whole test case is wrapped in the `rtos` namespace.

PR #489, opened 2019-05-29
Adding dependency injection for newlib

### LINKED ISSUE #81: Optimize uart0.min for newlib
To make newlib read and write more efficient, the uart0_putchar and getchar should be inlined. There should also be a uart0_send with a length parameter.

### LINKED ISSUE #401: Add dependency injection point for out() and in() functions in newlib.cpp
Currently uart0 is instantiated for LPC40xx platforms as the default out and in stream for communication. If we ever switch to USB implementation or something else, this global variable will be taking up space where its not needed. A more modular means of doing this is to dependency inject a function pointer for out() and in() during LowLevelInit() (soon to be named PlatformInitialize()).

kammce:
Resolves #401
Resolves #81

main_test.cpp now setups Stdout and Stdin for all tests, removing the
need for macros in the newlib.cpp file to destinguise between host tests
and firmware.

Added injection of Stdout and Stdin in both lpc40xx and lpc17xx.


kammce:
It turns out that issue #81 was not appropriate for this PR.

Issue #14, opened 2018-04-09
LPC40xx: EEPROM


kammce:

MasterBoyd:
Reserved for Harmander

kammce:
@harmandersihra is assigned to this project.

kammce:
Resolves by #575

PR #576, opened 2019-06-30
EEPROM Driver

harmandersihra:
Finished EEPROM driver with demo and unit tests.

Fixes issue #14


harmandersihra:
Changed the way write data is casted to avoid losing const-ness.

harmandersihra:
Rebased off of new master commits.

harmandersihra:
Was an issue where the date of the commit was 6 days ago, fixed that.

library/L1_Peripheral/lpc40xx/eeprom.hpp
```diff
@@ -0,0 +1,163 @@
...
+    // cause a hard fault, thus this safety is put into place to avoid that.
+    address &= 0b1111'1111'1100;
+
+    // Because the EEPROM uses 32-bit communication, read_data will be casted
+    // into a uint32_t *
+    uint32_t * read_data = reinterpret_cast<uint32_t *> (rdata);
+
+    uint16_t index = 0;
+
+    while ((index * 4) < count)
```
kammce: This is a for loop but in while loop form. Use a for loop.

library/L1_Peripheral/lpc40xx/test/eeprom_test.cpp
```diff
@@ -0,0 +1,92 @@
...
+                                  (rdata[2] << 16) +
+                                  (rdata[1] << 8) +
+                                  (rdata[0]);
+        CHECK(local_eeprom.ADDR == kAddress);
+        CHECK(local_eeprom.CMD == Eeprom::kRead32Bits);
+        CHECK(expected_value == kReadVal);
+
+        delete[] rdata;
+    }
+}
```
kammce: You must assign the `Eeprom::eeprom_register` back to its original value before the test. This is to put the object's global state variables back to what they were before the test, in the event that they are depended on in other tests.

library/L1_Peripheral/lpc40xx/test/eeprom_test.cpp
```diff
@@ -0,0 +1,92 @@
...
+        CHECK(local_eeprom.WDATA == kExpectedValue);
+
+        delete[] wdata;
+    }
+
+    SECTION("Reading")
+    {
+        constexpr uint16_t kAddress = 0x534;
+        constexpr uint32_t kReadVal = 0x12345678;
+        uint8_t * rdata = new uint8_t[4];
```
kammce: Why dynamically allocate when the size is known at compile time.
```suggestion
        uint8_t rdata[4];
```

library/L1_Peripheral/lpc40xx/test/eeprom_test.cpp
```diff
@@ -0,0 +1,92 @@
...
+
+        uint32_t expected_value = (rdata[3] << 24) +
+                                  (rdata[2] << 16) +
+                                  (rdata[1] << 8) +
+                                  (rdata[0]);
+        CHECK(local_eeprom.ADDR == kAddress);
+        CHECK(local_eeprom.CMD == Eeprom::kRead32Bits);
+        CHECK(expected_value == kReadVal);
+
+        delete[] rdata;
```
kammce: This isn't necessary if you change rdata to be stack allocated.

library/L1_Peripheral/lpc40xx/test/eeprom_test.cpp
```diff
@@ -0,0 +1,92 @@
...
+    Eeprom test_eeprom(mock_system_controller.get());
+
+    // Write/Read tests only check one 32-bit transfer, so size is 4 bytes
+    constexpr size_t kPayloadSize = 4;
+
+    SECTION("Initialization")
+    {
+        test_eeprom.Initialize();
+
+        // Assuming clock speed of 48 Mhz
```
kammce: No need to assume, you defined this for your test above.
```suggestion
```

library/L1_Peripheral/lpc40xx/test/eeprom_test.cpp
```diff
@@ -0,0 +1,92 @@
...
+        // Wait State Register Values should be 2, 3, and 1
+        constexpr uint32_t kWaitStateValues = 0b00000010'00000011'00000001;
+        // Clock divider should be set to 128
+        constexpr uint32_t kClockDivider    = 0b10000000;
```
kammce:
```suggestion
        constexpr uint32_t kClockDivider    = 128;
```
Just making the expected clock divider 128 makes this more readable, rather than having it in binary here. It also removes the need for the comment below.

library/L1_Peripheral/lpc40xx/test/eeprom_test.cpp
```diff
@@ -0,0 +1,92 @@
...
+    constexpr size_t kPayloadSize = 4;
+
+    SECTION("Initialization")
+    {
+        test_eeprom.Initialize();
+
+        // Assuming clock speed of 48 Mhz
+        // Wait State Register Values should be 2, 3, and 1
+        constexpr uint32_t kWaitStateValues = 0b00000010'00000011'00000001;
+        // Clock divider should be set to 128
```
kammce:
```suggestion
```
See comment below.

library/L1_Peripheral/lpc40xx/test/eeprom_test.cpp
```diff
@@ -0,0 +1,92 @@
...
+        constexpr uint32_t kClockDivider    = 0b10000000;
+
+        CHECK(local_eeprom.WSTATE == kWaitStateValues);
+        CHECK(local_eeprom.CLKDIV == kClockDivider);
+    }
+    SECTION("Writing")
+    {
+        uint8_t * wdata = new uint8_t[4];
```
kammce: Refrain from using heap when you don't need to. See comments in "Reading" section.

demos/sjtwo/eeprom/source/main.cpp
```diff
@@ -0,0 +1,50 @@
...
+#include "utility/log.hpp"
+#include "utility/time.hpp"
+
+int main(void)
+{
+  LOG_INFO("Starting EEPROM Example");
+  constexpr size_t kPayloadSize = 128;
+  constexpr uint32_t kAddress   = 0b0110'0011'1000;
+
+  uint8_t * results = new uint8_t[kPayloadSize];
```
kammce: https://sjsu-dev2.readthedocs.io/en/latest/rst/contributing/style.html#memory-allocation

Refrain from using dynamic memory.
```suggestion
  uint8_t results[kPayloadSize];
```

demos/sjtwo/eeprom/source/main.cpp
```diff
@@ -0,0 +1,50 @@
...
+
+  LOG_INFO("Initializing EEPROM");
+  sjsu::lpc40xx::Eeprom eeprom;
+  eeprom.Initialize();
+
+  // Generating Payload
+  uint8_t * list = new uint8_t[kPayloadSize];
```
kammce: Same as above.

demos/sjtwo/eeprom/source/main.cpp
```diff
@@ -0,0 +1,50 @@
...
+
+  for (size_t i = 0; i < kPayloadSize; i++)
```
kammce: Its more direct and readable if you check the size of the array like so. But it requires that the size be knowable at compile time, which requires that you move away from heap allocation.
```suggestion
  for (size_t i = 0; i < std::size(list); i++)
```

demos/sjtwo/eeprom/source/main.cpp
```diff
@@ -0,0 +1,50 @@
...
+    eeprom.Read(results, kAddress, kPayloadSize);
+
+    LOG_INFO("Results:");
+    for (size_t i = 0; i < kPayloadSize; i++)
+    {
+      LOG_INFO("%i: %u", i, results[i]);
+    }
+    LOG_INFO("");
+
+    sjsu::Delay(1000);
```
kammce: Oh god! This will destroy the boards EEPROM if you leave this program running for a while. Maybe make this interactive so that the user has to press a button in order to write and read from the device. Basically pass the buck to the user of the board. Probably add a log message stating that the too many writes will cause the eeprom to fail.

demos/sjtwo/eeprom/source/main.cpp
```diff
@@ -0,0 +1,50 @@
...
+
+    LOG_INFO("Starting Read");
+    eeprom.Read(results, kAddress, kPayloadSize);
+
+    LOG_INFO("Results:");
+    for (size_t i = 0; i < kPayloadSize; i++)
+    {
+      LOG_INFO("%i: %u", i, results[i]);
+    }
+    LOG_INFO("");
```
kammce:
```suggestion
    printf("\n");
```

demos/sjtwo/eeprom/source/main.cpp
```diff
@@ -0,0 +1,50 @@
+#include <project_config.hpp>
+
+#include <cstdint>
+#include <iterator>
+
+#include "L1_Peripheral/lpc40xx/eeprom.hpp"
+#include "L2_HAL/displays/led/onboard_led.hpp"
```
kammce: Are you using this?

library/L1_Peripheral/lpc40xx/eeprom.hpp
```diff
@@ -0,0 +1,163 @@
...
+    {
+      eeprom_register->ADDR = address;
```
kammce: The address can be computed using the index here, which allows this to be a for loop.
```suggestion
      eeprom_register->ADDR = address + (index * 4);
```

library/L1_Peripheral/lpc40xx/eeprom.hpp
```diff
@@ -0,0 +1,163 @@
...
+      eeprom_register->CMD  = kRead32Bits;
+      read_data[index]      = eeprom_register->RDATA;
+
+      address += 4;
```
kammce:
```suggestion
```

library/L1_Peripheral/lpc40xx/eeprom.hpp
```diff
@@ -0,0 +1,163 @@
...
+  // object is used (kLpc40xxSystemController)
+  explicit constexpr Eeprom(const sjsu::SystemController & system_controller =
+                             kLpc40xxSystemController)
+      : system_controller_(system_controller)
+  {
+  }
+
+  // Initializing the EEPROM requires setting the wait state register,
+  // and the clock divider register
+  void Initialize() const override
```
kammce: Does the EEPROM need to be powered on?

kammce:
Awesome job Harmander. This commit looks fantastic. But there are a few things you need to address first before this can be merged.

library/L1_Peripheral/lpc40xx/eeprom.hpp
harmandersihra: The EEPROM is powered on by default on startup, but it can be powered down to save power.

-Table 731 from UM10562.pdf

library/L1_Peripheral/lpc40xx/eeprom.hpp
```diff
@@ -0,0 +1,163 @@
...
+  {
+    const float kCclk           = static_cast<float>
+                                    (system_controller_.GetSystemFrequency());
```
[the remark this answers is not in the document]
harmandersihra: It uses the CPU clock, according to sections 37.5.1.5/6 in UM10562.pdf

demos/sjtwo/eeprom/source/main.cpp
harmandersihra: That's something I hadn't considered, but the button thing is a great idea. I currently don't have an SJ2 with buttons (only have the small one right now) so I can't test this yet. I'll try to get a hold of one and add this in.

library/L1_Peripheral/lpc17xx/eeprom.hpp
```diff
@@ -0,0 +1,12 @@
+#pragma once
```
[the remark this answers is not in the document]
harmandersihra: Does this mean just delete the file? That's what it sounded like, but if that's not the right thing to do, I'll fix it.

harmandersihra:
The button fix still has to be added in, but everything else has been addressed. I'm not sure if you want to hold back on reviewing this until I've added that in or not. Also, if you want me to delete this PR for now, let me know. I'm not entirely sure when I'll be able to get a full size SJ2, but hopefully within a week.

library/L1_Peripheral/lpc40xx/eeprom.hpp
kammce: Hmmm, run the power on anyway, in the event someone has powered it down for some reason, but then wants to power it back up later.

library/L1_Peripheral/lpc40xx/eeprom.hpp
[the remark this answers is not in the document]
kammce: :+1:

library/L1_Peripheral/lpc40xx/eeprom.hpp
```diff
@@ -0,0 +1,155 @@
...
+      // Page offset is incremented by 4 because we're writing 32 bits
+      constexpr uint8_t kOffsetInterval = 4;
+
+      address = static_cast<uint16_t>((page_count << 6) + page_offset);
+
+      eeprom_register->ADDR  = address;
+      eeprom_register->CMD   = kWrite32Bits;
+      eeprom_register->WDATA = write_data[0];
+
+      write_data++;
```
kammce: Refrain from doing C style pointer arithmetic. This can be easily solved by using bracket notation:

`write_data[i]` rather than `write_data[0]` each time.

library/L1_Peripheral/lpc40xx/eeprom.hpp
```diff
@@ -0,0 +1,155 @@
...
+    const uint32_t * write_data = reinterpret_cast<const uint32_t *>(wdata);
+
+    // The first 6 bits in the address (MSB) dictate which page is being written
+    // to in the EEPROM, and the last 6 bits (LSB) dictate offset in the page
+    uint8_t page_count  = static_cast<uint8_t>(full_address >> 6);
+    uint8_t page_offset = full_address & 0b0000'0011'1111;
+
+    uint16_t address;
+
+    for (; count > 0; count -= 4)
```
kammce:
```suggestion
    for (int i = 0; i < count; i += 4)
```
Once you make this change, you can use `i` in places where you are using `[]`

library/L1_Peripheral/lpc40xx/eeprom.hpp
```diff
@@ -0,0 +1,155 @@
...
+
+  // The EEPROM is accessed through a 64 byte page buffer, and after it fills
+  // up, it must be programmed to the actual EEPROM. This function handles that.
+  void Program(uint32_t address) const
+  {
+    eeprom_register->ADDR  = address;
+    eeprom_register->CMD  = kEraseProgram;
+
+    // Poll status register bit to see when programming is finished
+    Wait((!(eeprom_register->INT_STATUS >> kProgramStatusBit)) & 1);
```
kammce: You are going to need to turn this into a lambda for this to work.

library/L1_Peripheral/lpc40xx/eeprom.hpp
```diff
@@ -0,0 +1,155 @@
...
+  // This function will return however much 32-bit data from the EEPROM
+  // starting at address and continuing on for how large count is. The EEPROM
+  // can only be interfaced in 32-bit mode, which means that the address used
+  // must be a multiple of 4. The address entered will be cut down to a
+  // multiple of 4.
+  void Read(uint8_t * rdata, uint32_t address, size_t count) const override
+  {
+    // Writing a 32-bit number to an address that isn't a multiple of 4 will
+    // cause a hard fault, thus this safety is put into place to avoid that.
+    address &= 0b1111'1111'1100;
```
kammce: Since this mask is used a lot, this should become a `static constexpr` member of this class, less you misspell it at some point.

library/L1_Peripheral/lpc40xx/eeprom.hpp
```diff
@@ -0,0 +1,155 @@
...
+
+    // Because the EEPROM uses 32-bit communication, read_data will be casted
+    // into a uint32_t *
+    uint32_t * read_data = reinterpret_cast<uint32_t *> (rdata);
```
kammce: Remove extra space
```suggestion
    uint32_t * read_data = reinterpret_cast<uint32_t *>(rdata);
```

library/L1_Peripheral/lpc40xx/eeprom.hpp
harmandersihra:
```suggestion
    for (size_t i = 0; i*4 < count; i++)
```

I was thinking of doing this, since count is the number of uint8 values, but write_data is uint32. This allows me to just do write_data[i] instead of write_data[i/4]. Thoughts?

library/L1_Peripheral/lpc40xx/eeprom.hpp
harmandersihra: The Wait statement? What do you mean by work? It seems to be working right now.

library/L1_Peripheral/lpc40xx/eeprom.hpp
kammce: Your idea to do `i*4 < count`sounds fine to me. Go for it.

harmandersihra:
This should be the final driver with button control for restarting the demo, and lambdas function with a proper timeout for the Wait() functions.

harmandersihra:
Fixed rebase issue.

PR #593, opened 2019-07-09
Fix typo which causes SD card status responses to fail

rjavier441:
Fixes a typo that prevents SD Card communication, and removes delays internal to the driver. Manual testing with the removed delays has shown to have no negative impact to the SD card's usage (tested using SD Card Demo)

### DIFF
--- a/library/utility/crc.hpp
+++ b/library/utility/crc.hpp
@@ -1,0 +1,100 @@
+#pragma once
+
+#include <cstdlib>
+#include <cstdint>
+#include <type_traits>
+
+namespace sjsu
+{
+namespace crc
+{
+/// @description     A template housing a CRC7 or CRC16 table for the SD
+///                  card, using either an 8-bit or 16-bit data type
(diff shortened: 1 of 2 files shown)


kammce:
Please squash your commits.

library/utility/crc.hpp
```diff
@@ -0,0 +1,99 @@
+#pragma once
+
+#include <cstdlib>
+#include <cstdint>
```
kammce: https://en.cppreference.com/w/cpp/types/is_same

You didn't include <type_traits> (the thing that has std::is_same in it) here. Tidy has caught that the file on its own does not include anything. You get lucky because in the other files, type_traits is probably included somewhere in there.

rjavier441: Alright, I'll do this tonight

rjavier441:
@kammce Can you review my change? I believe it has been resolved.

PR #612, opened 2019-07-17
Make flash target platform defined

### LINKED ISSUE #606: Make flash target platform defined
Within each platform's .mk file, will be a definition of its `flash` target. This way, each platform can specify how it is supposed to be flashed.

kammce:
Resolves #606


library/L0_Platform/lpc17xx/lpc17xx.mk
```diff
@@ -16,3 +16,10 @@ OPENOCD_CONFIG = $(LIBRARY_DIR)/L0_Platform/lpc17xx/lpc17xx.cfg
 $(eval $(call BUILD_LIRBARY,liblpc17xx,LIBRARY_LPC40XX))
 
 include $(LIBRARY_DIR)/L0_Platform/arm_cortex/m3/m3.mk
+
+flash:
```
kammce: rename this to `platform-flash`. Then add `flash` back to the main `makefile`. The idea is to allow the main flash target to cause a `make application` then a `make platform-flash`, this way, we no longer need to include `@$(MAKE) --quiet application` in every platform's .mk file.

PR #626, opened 2019-07-21
Adding LPC17xx L1 configuration structures

HuangDave:
- Added configuration structures for the following peripherals:
  - ADC
  - DAC
  - PWM
  - Timer

Resolves: #561, #562, #566, #567

### DIFF
--- a/library/L1_Peripheral/lpc17xx/adc.hpp
+++ b/library/L1_Peripheral/lpc17xx/adc.hpp
@@ -1,5 +1,6 @@
 #pragma once
 
+#include "L1_Peripheral/lpc17xx/pin.hpp"
 #include "L1_Peripheral/lpc40xx/adc.hpp"
 
 namespace sjsu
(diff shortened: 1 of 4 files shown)


library/L1_Peripheral/lpc17xx/spi_bus.hpp
```diff
@@ -0,0 +1,37 @@
+#pragma once
+
+#include "L1_Peripheral/lpc17xx/pin.hpp"
+#include "L1_Peripheral/lpc17xx/spi.hpp"
```
kammce: Which one would I include in order to use SPI? As it looks here, I have to `#include "spi_bus.hpp"` if that is the case, it would be better to swap names, having this become `spi.hpp` and have the other be `spi_bus.hpp`

library/L1_Peripheral/lpc17xx/i2c_bus.hpp
```diff
@@ -0,0 +1,69 @@
+#pragma once
+
+#include "L1_Peripheral/lpc17xx/pin.hpp"
+#include "L1_Peripheral/lpc17xx/i2c.hpp"
```
kammce: see comment mentioned at `spi_bus.hpp`

library/L1_Peripheral/lpc17xx/spi_bus.hpp
HuangDave: To use SPI, `spi_bus.hpp` should be included. For some reason swapping the names would cause the errors to show up.

kammce: Ok, for now, lets pull in all of the changes that do not require the `_bus.hpp` hack to work. We can then spend some time to figure out why the naming is so fragile. Because allowing this problem to stay this will cause more headache down the road.

kammce:
Please remove `spi_bus` and `i2c_bus` before you merge your stuff in.

PR #630, opened 2019-07-23
Move away from nested namespace inclusion

### LINKED ISSUE #629: Move from wrapping include in namespace
Moving away from the current L0 standard, which is to nest/wrap the inclusion of the platform files within a namespace. This helped in the short term but become a massive problem when you want something to be multi-platform and you end up including to version of a similar with names that collide with eachother. 

This isn't desirable because it means that we need to edit these files each time we add a new platform, but its better than the previous hack with respect to issues that could occur down the line.

kammce:
Resolves #629


kammce:
Merged in #631

PR #659, opened 2019-08-06
Adding Temt6000x01 Ambient Light Sensor Driver

HuangDave:
- Added the Light interface to L2_HAL/sensors/enviornment.
- Added the Temt6000x01 ambient light sensor driver.
- Enabled the following units
    - illuminance
    - luminous flux (required for illuminance)
    - voltage
- Fixed the calculation of kPinMask in the constructor of lpc17xx::Pin
  to wrap around to 0 when the position exceeds the max pin number of 32.


library/L2_HAL/sensors/environment/light.hpp
```diff
@@ -0,0 +1,14 @@
+#pragma once
+
+#include "utility/units.hpp"
+
+namespace sjsu
+{
+class Light
```
kammce:
```suggestion
class LightSensor
```
Because Light makes it sound like we are controlling a lightbulb.

library/L2_HAL/sensors/environment/light.hpp
```diff
@@ -0,0 +1,14 @@
...
+{
+ public:
+  virtual bool Initialize() const                          = 0;
+  virtual units::illuminance::lux_t GetIlluminance() const = 0;
+  virtual float GetIlluminancePercentage() const           = 0;
```
kammce:
```suggestion
  virtual float GetPercentageBrightness() const           = 0;
  virtual float GetMaxIlluminance() const           = 0;
```
I think giving the maximum illuminance would be a better as it allows the developer using this class the option of what they want to do with it, for example, maybe they are looking at the saturation value of the sensor because they have an array of them and some of them have better accuracy at brighter environments then in dark environments. Knowing the maximum lux that it can read can help to know when to switch to the other sensor.

I also think that **Brightness** sounds better then Illuminance with regards to percentage.

library/L2_HAL/sensors/environment/light/temt6000x01.hpp
```diff
@@ -0,0 +1,53 @@
...
+#include "L1_Peripheral/adc.hpp"
+#include "utility/log.hpp"
+
+namespace sjsu
+{
+class Temt6000x01 final : public Light
+{
+ public:
+  /// Pull-down resistor
+  static constexpr uint32_t kResistance = 10'000;
```
kammce: This is not a constant in the datasheet, meaning other developers may use a different resistor. We shouldn't assume everyone is using 10k. This should be a constructor parameter.

library/L2_HAL/sensors/environment/light/temt6000x01.hpp
```diff
@@ -0,0 +1,53 @@
...
+  /// 12-bit ADC resoltion = 2^12 - 1 = 4'095
+  static constexpr uint32_t kResolution = 4'095;
```
kammce: This is platform dependent. Use Adc::GetActiveBits() to get the resolution of the passed adc object.

library/L2_HAL/sensors/environment/light/temt6000x01.hpp
```diff
@@ -0,0 +1,53 @@
...
+  /// ADC Reference voltage.
+  static constexpr units::voltage::microvolt_t kReferenceVoltage = 3.3_V;
+  /// The conversion ratio to convert from the ADC output to
+  /// illuminance.
+  ///
+  /// From Fig. 3 - Photo Current vs. Illuminance of TEMT6000X01 datasheet,
+  /// lux = 2 * I, where current, I, is in micro-amps
+  ///     = 2 * adc_out * ((Vref / resolution) / resistance)
+  static constexpr float kConversionRatio =
```
kammce: The ADC -> Lux equation is not linear. If you check the datasheet you will see that the graph is logarithmic. We must convert the graph in the datasheet to an equation and use that.

kammce:
Great work man!

PR #672, opened 2019-08-10
Added LPC40xx::Watchdog Driver

### LINKED ISSUE #12: LPC40xx: Watchdog

aidenchan12:
Resolves #12


aidenchan12:
This commit is for LPC40xx: Watchdog (L1).

kammce:
You need to incorporate a line that says `Resolves #12` in your commit message. Please amend and force push that.

library/L1_Peripheral/lpc40xx/watchdog_interface.hpp
```diff
@@ -0,0 +1,19 @@
...
+#include <cstdint>
+#include <iterator>
+
+namespace sjsu
+{
+
+class WatchdogInterface
+{
+ public:
+     virtual void Initialize(uint32_t timer_constant, uint32_t timer_window,
```
kammce: Also not all watch dogs will be WINDOWED watch dogs. So this may be too specific to the LPC40xx use case. What do other watchdog drivers look like? You may want to check out some other libraries like Keil and MBED for some examples. Then come back and reformulate this interface.

library/L1_Peripheral/lpc40xx/watchdog_interface.hpp
```diff
@@ -0,0 +1,19 @@
...
+     uint32_t timer_warning) = 0;
```
kammce:
```suggestion
     uint32_t timer_warning) const = 0;
```
All L1 peripheral interface methods should be made const.

library/L1_Peripheral/lpc40xx/watchdog_interface.hpp
```diff
@@ -0,0 +1,19 @@
...
+     virtual void RegisterIrqHandler() = 0;
```
kammce: I feel like this shouldn't be apart of the interface and should only be found in the implementation. Thus, this should be removed.

library/L1_Peripheral/lpc40xx/test/watchdog_test.cpp
```diff
@@ -0,0 +1,54 @@
...
+    // Clear local wdt registers
+    memset(&local_wdt, 0, sizeof(local_wdt));
+
+    // Set base registers to local registers
+    Watchdog::wdt_base = &local_wdt;
+
+    Watchdog watchdog_test;
+    watchdog_test.Initialize(kTimerConstant, kTimerWindow, kTimerWarning);
+    watchdog_test.FeedSequence();
+    watchdog_test.CheckTimeout();
```
kammce: Move `CheckTimeout()` call this to its test section.

library/L1_Peripheral/lpc40xx/test/watchdog_test.cpp
```diff
@@ -0,0 +1,54 @@
...
+        CHECK(kTimerWarning == local_wdt.WARNINT);
+    }
+
+    SECTION("Send Feed Sequence")
+    {
+        CHECK(kFeedSequence == local_wdt.FEED);
+    }
+    SECTION("Read watchdog timer value")
+    {
+        CHECK(kTimerValue == local_wdt.TV);
```
kammce: This test does not make much sense. Yes, you test that the value is zero, but this doesn't exercise your watch dog code whats so ever. You should actually do something like this:

```suggestion
        constexpr uint32_t kNewTimerValue = 150;
        local.wdt.TV = kNewTimerValue;
        CHECK(kNewTimerValue == watchdog_test.CheckTimeout());
```

This way, you check that the correct TV value is being returned.

library/L1_Peripheral/lpc40xx/test/watchdog_test.cpp
```diff
@@ -0,0 +1,54 @@
...
+    LPC_WDT_TypeDef local_wdt;
+
+    // Clear local wdt registers
+    memset(&local_wdt, 0, sizeof(local_wdt));
+
+    // Set base registers to local registers
+    Watchdog::wdt_base = &local_wdt;
+
+    Watchdog watchdog_test;
+    watchdog_test.Initialize(kTimerConstant, kTimerWindow, kTimerWarning);
```
kammce: Move `Initialize()` call this to its test section.

library/L1_Peripheral/lpc40xx/test/watchdog_test.cpp
```diff
@@ -0,0 +1,54 @@
...
+    watchdog_test.FeedSequence();
```
kammce: Move `FeedSequence()` call this to its test section.

library/L1_Peripheral/example/watchdog.cpp
```diff
@@ -0,0 +1,61 @@
...
+    watchdog.FeedSequence();
+  }
+}
+
+void HighCPUTask(void * pv_paramater)
+{
+  while (true)
+  {
+     printf("High Task! \n");
+     sjsu::Delay(500ms);
```
kammce: Put a note here as to why you are using `sjsu::Delay()` rather than vTaskDelay for the readers of the example. They need to know that you are doing this because you want to starve the tasks below you to cause the watchdog to be asserted if the delay is too long.

library/L1_Peripheral/example/watchdog.cpp
```diff
@@ -0,0 +1,61 @@
...
+    if (button3.Read()) {
+      led3.SetLow();
+    }
+    else {
+      led3.SetHigh();
```
kammce: Does anything happen in this example? Wouldn't it be cool if you had the button cause a `WatchDog` interrupt by having the button cause an additional delay that goes beyond the time of the watchdog timer. Be creative with this. And add a bunch of notes so the readers know whats happening.

library/L1_Peripheral/lpc40xx/watchdog_interface.hpp
kammce: Actually! What would necessary is that we need to be able to assign our own handler for the WatchDog. So when the WatchDog interrupt is triggered, our function gets called.

library/L1_Peripheral/lpc40xx/watchdog_interface.hpp
```diff
@@ -0,0 +1,19 @@
...
+     virtual void FeedSequence() = 0;
+     virtual uint32_t CheckTimeout() = 0;
+};
```
kammce: Maybe too much for this commit, (meaning do not implement this now), but it would be great if we could have the watchdog reset the board it is triggered.

kammce:
@Aiden Chan please squash your commits and update branch.

kammce:
Change the name of the PR and the commit to something that makes more sense then Resolves #___.

demos/sjone/light_sensor/makefile
```diff
@@ -0,0 +1,15 @@
+# sjsu_dev2.mk holds the $(SJSU_DEV2_BASE) variable which holds the location of
```
kammce: What is light sensor code doing in here?

aidenchan12: I might have done something wrong when squashing my commits

library/L1_Peripheral/lpc40xx/watchdog.hpp
```diff
@@ -0,0 +1,89 @@
...
+    // LED0 will flash upon entering the interrupt handler
+    Gpio led0(2, 3);
+    led0.SetDirection(sjsu::Gpio::Direction::kOutput);
+    led0.SetLow();
+  }
+
+  // This function converts seconds to units used by the watchdog
+  static uint32_t ConvertSeconds(uint32_t timer_variable)
+  {
+    // Every 125000 units written to the TC register is equal to 1 second
```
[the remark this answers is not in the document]
aidenchan12: In the datasheet it says that by it uses the watchdog clock source (Twdclk) with a frequency of 500 kHz. It says that the formula is: (Twdclk)*(Value written to the register)*(4). I was able to come up with 125000 units = 1 second by recording the amount of time it took for the watchdog interrupt to trigger. I tested it by passing in 10, 20, and 60 (seconds) into the function. I am still not sure what is the correct way to convert the value written to the register into seconds.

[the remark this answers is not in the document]
kammce: Oh neat. So it does have its own clock source. It might be good to rename this from `ConvertSeconds()` (what does that even mean), to `ConvertDurationToWatchDogTicks()`. This gives you a better idea of what it is you are doing when you run this method.

You could probably do something like:
```suggestion
   static uint32_t ConvertDurationToWatchDogTicks(std::chrono::microseconds duration)
   { 
     // Every 125000 units written to the TC register is equal to 1 second
     return (duration.count() * 500_kHz) / 4;
   }
```

My suggestion above isn't the exact thing you should do. It won't compile. But you get the idea from looking at it. That's all you would have to do really.

kammce:
Please fix up your commit as it seems that other older commits are showing up in this PR.

library/L1_Peripheral/lpc40xx/watchdog.hpp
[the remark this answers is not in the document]
aidenchan12: Ok, I renamed the method. I did something similar to that example, but now I am having trouble converting the std::chrono::microseconds object into a uint32_t type to write the TC register.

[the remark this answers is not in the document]
kammce: Here is a sample:
```
std::chrono::microseconds duration = 10us;
duration.count();  // returns value 10.
```

PR #675, opened 2019-08-15
Adding Tmp102 Temperature Sensor

HuangDave:
- Fixed lpc17xx::I2cBus::kI2c2SclPin to have the correct pin number.

Resolves: #532

### DIFF
--- a/library/L2_HAL/sensors/environment/temperature/test/tmp102_test.cpp
+++ b/library/L2_HAL/sensors/environment/temperature/test/tmp102_test.cpp
@@ -1,0 +1,134 @@
+#include "L1_Peripheral/system_controller.hpp"
+#include "L2_HAL/sensors/environment/temperature/tmp102.hpp"
+#include "L4_Testing/testing_frameworks.hpp"
+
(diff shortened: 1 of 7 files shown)


library/L2_HAL/sensors/environment/temperature/test/tmp102_test.cpp
```diff
@@ -0,0 +1,122 @@
...
+  constexpr uint8_t kDeviceAddress = Tmp102::DeviceAddress::kGround;
+  Mock<I2c> mock_i2c;
+  Tmp102 temperature_sensor(mock_i2c.get(), kDeviceAddress);
+
+  When(Method(mock_i2c, Initialize)).AlwaysReturn(Status::kSuccess);
+
+  SECTION("Initialize")
+  {
+    const Status kInitializationStatus = temperature_sensor.Initialize();
+    CHECK(kInitializationStatus == Status::kSuccess);
```
kammce: It maybe useful to check that your driver actually passes a non-success status back. So if init did fail we have a check for that here.

PR #676, opened 2019-08-15
Add Stop Watch library and test

### LINKED ISSUE #34: Stop Watch
Stopwatch implementation where the user can get the time between executions.

jechea14:
Resolves #34

### DIFF
--- a/library/utility/stopwatch.hpp
+++ b/library/utility/stopwatch.hpp
@@ -1,0 +1,54 @@
+// stopwatch.hpp Overview
+#pragma once
+#include <cstdint>
+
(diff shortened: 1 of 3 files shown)


library/utility/stopwatch.hpp
```diff
@@ -0,0 +1,54 @@
+// stopwatch.hpp Overview
+#pragma once
+#include <cstdint>
+
+namespace sjsu
+{
+class StopWatch
+{
+ public:
+  enum class DeltaScale
```
kammce: This should probably be removed once we use the units library.

PR #686, opened 2019-08-27
Update example to show Write mocking

kammce:


kammce:
Discovered that the test already includes a way to mock writing.

PR #688, opened 2019-08-30
Added capture feature to library

### LINKED ISSUE #505: LPC40xx: Pulse Capture

charlesmacd:
Pull request with squashed history. Closes #505.

### DIFF
--- a/demos/multiplatform/dual_platforms/source/main.cpp
+++ b/demos/multiplatform/dual_platforms/source/main.cpp
@@ -35,8 +35,8 @@ int main()
   }
   else
   {
-    button_gpio = &sjsu::GetInactive<sjsu::Gpio>();
-    led_gpio    = &sjsu::GetInactive<sjsu::Gpio>();
+    LOG_CRITICAL("Invalid platform for this application!");
+    sjsu::Halt();
   }
   // Phase #3:
   // Pass and initialize all L2 HAL abstraction objects. Typically the hardware
@@ -45,7 +45,7 @@ int main()
   // application.
   sjsu::Button button(*button_gpio);
 
-  // Phase #4:
+  // DONE!
   // Now you can use the L2 HAL abstractions and L1 Peripheral interfaces as
   // you wish.
 
--- a/documentation/guides/multi_platform.md
+++ b/documentation/guides/multi_platform.md
@@ -7,5 +7,95 @@ implementation details from each platform you want to support.
 
 See the example code `demos/multiplatform/dual_platforms/`.
 
-The method for making a project multi-platform has 4 phases to it, see file for
+The method for making a project multi-platform has 3 phases to it, see file for
 more details.
+
+## Phase 1: Create interface pointers
+In this phase you will want to define all of the peripherals you plan to use as
+pointers. Typically, references are prefered for interfaces, but references
+cannot be re-assigned at runtime, thus we must use pointers.
+
+``` C++
+const sjsu::Gpio * status_led_gpio;
+const sjsu::Gpio * power_on_gpio;
+const sjsu::Gpio * kill_switch_gpio;
+
+const sjsu::Spi * spi_bus;
+const sjsu::I2c * i2c_bus;
+
+const sjsu::Gpio * motor_direction_gpio;
+const sjsu::Pwm * motor_controller_pwm;
+```
+
+## Phase 2: Assign pointers in `if constepxr` blocks per platform
+Now that we have made our list of peripherals or application needs, we need to
+assign these pointers to real peripheral driver objects for the specified
+platform.
+
+There will need to be `if constexpr` block for each platform you care to
+support.
+
+Within each block a `static` peripheral object will need to be created and
+constructed for each of the pointers in phase 1. After construction, each
+peripheral pointer will need to be assigned to the address of each `static`
+peripheral object. In this phase you can also create temperary objects within
+the block if this is found useful.
+
+The `static` keyword is required for objects that have their addresses
+referenced outside the scope of the `if constepxr` block. Using the `static`
+keyword will allocate the object statically and will make the object available
+for the whole lifetime of the application. Not doing this means that the objects
+will be destructed after the program leaves the scope of the `if constexpr`
+block. If this happens, the behavior of the program is undefined.
+
+``` C++
+if constexpr (sjsu::build::kPlatform == sjsu::build::Platform::lpc17xx)
+{
+  static sjsu::lpc17xx::Gpio lpc17xx_status_led_gpio(1, 9);
+  static sjsu::lpc17xx::Gpio lpc17xx_poweron_gpio(1, 0);
+  // etc...
+
+  poweron_gpio    = &lpc17xx_poweron_gpio;
+  status_led_gpio = &lpc17xx_status_led_gpio;
+  // etc...
+}
+else if constexpr (sjsu::build::kPlatform == sjsu::build::Platform::lpc40xx)
+{
+  static sjsu::lpc17xx::Gpio lpc40xx_status_led_gpio(1, 9);
+  static sjsu::lpc17xx::Gpio lpc40xx_poweron_gpio(1, 0);
+  // etc...
+
+  poweron_gpio    = &lpc40xx_poweron_gpio;
+  status_led_gpio = &lpc40xx_status_led_gpio;
+  // etc...
+}
+else if constexpr (sjsu::build::kPlatform == sjsu::build::Platform::linux)
+{
+  static sjsu::lpc17xx::Gpio linux_status_led_gpio(1, 9);
+  static sjsu::lpc17xx::Gpio linux_poweron_gpio(1, 0);
+  // etc...
+
+  poweron_gpio    = &linux_poweron_gpio;
+  status_led_gpio = &linux_status_led_gpio;
+  // etc...
+}
+else
+{
+  LOG_CRITICAL("Invalid platform for this application!");
+  sjsu::Halt();
+}
+```
+
+## Phase 3: Pass references to L2_HAL and L3_Applications objects if any
+If you are passing the L1 peripherals to L2 HALS then this phase is the time to
+do it. This phase is outside of the `if constexpr` block and is where you list
+and construct every hardware abstraction. Same goes for application level
+objects. A small example below
+
+``` C++
+sjsu::Button poweron_button(*poweron_gpio);
+sjsu::RCCarController rc_controller(*motor_direction_gpio, *motor_controller_pwm, /* etc ... */);
+```
+
+## Done
+At this point you can begin using your peripherals, HALs, application layer.


library/L1_Peripheral/capture.hpp
```diff
@@ -0,0 +1,32 @@
+#pragma once
+#include <cstdint>
+#include "L1_Peripheral/interrupt.hpp"
```
kammce:
```suggestion

#include "L1_Peripheral/interrupt.hpp"
```

library/L1_Peripheral/capture.hpp
```diff
@@ -0,0 +1,32 @@
+#pragma once
```
kammce:
```suggestion
#pragma once

```

library/L1_Peripheral/capture.hpp
```diff
@@ -0,0 +1,32 @@
...
+#include <cstdint>
+#include "L1_Peripheral/interrupt.hpp"
+#include "utility/status.hpp"
+#include "utility/units.hpp"
+
+namespace sjsu {
```
kammce: run clang format on this ( `make format-code` )

library/L1_Peripheral/lpc40xx/capture.hpp
```diff
@@ -0,0 +1,195 @@
...
+#include "L1_Peripheral/lpc40xx/system_controller.hpp"
+#include "L1_Peripheral/lpc40xx/timer.hpp"
+#include "L1_Peripheral/lpc40xx/capture.hpp"
+#include "L1_Peripheral/timer.hpp"
+#include "L1_Peripheral/capture.hpp"
+#include "utility/log.hpp"
+#include "utility/status.hpp"
+#include "utility/units.hpp"
+
+
```
kammce: Remove extra line

library/L1_Peripheral/lpc40xx/capture.hpp
```diff
@@ -0,0 +1,195 @@
+#pragma once
+#include <cstdint>
```
kammce: Add newline between `pragma` `c++ libraries` and `sjsu-dev2` libraries, throughout all of the code.

library/L1_Peripheral/lpc40xx/capture.hpp
```diff
@@ -0,0 +1,195 @@
...
+
+    struct Channel // NOLINT
+    {
+     private:
+        inline static IsrPointer timer0_isr                 = nullptr;
+        inline static const Timer::ChannelPartial_t kTimerPartial0 = {
+        .timer_register = LPC_TIM0,
+        .power_id       = SystemController::Peripherals::kTimer0,
+        .irq            = IRQn::TIMER0_IRQn,
+        .user_callback  = &timer0_isr,
```
kammce: If it is feasible, it would be nice to include a pin parameter here with the information needed to setup a pin for capture. We don't need to be able to control every pin, just a few. If the developer wants to use a pin we don't have they can construct their own Channel_t object.

library/L1_Peripheral/lpc40xx/capture.hpp
```diff
@@ -0,0 +1,195 @@
...
+    static void TimerHandler(const Timer::ChannelPartial_t & channel)
+    {
+        if (*channel.user_callback != nullptr)
+        {
+            (*channel.user_callback)();
+        }
+        channel.timer_register->IR |= 0x30;
+    }
+
+    typedef struct
```
kammce: Do not use `typedef` in C++. Prefer using Using. Also, there is no need to typedef structs in C++ at all. You can simply just use `struct` like you would `class`.

library/L1_Peripheral/lpc40xx/capture.hpp
```diff
@@ -0,0 +1,195 @@
...
+    {
+        uint32_t count;
+        uint32_t flags;
+    } CaptureStatus_t;
```
kammce: Prefer to not define structs at the end like this.

library/L1_Peripheral/lpc40xx/capture.hpp
```diff
@@ -0,0 +1,195 @@
...
+    };
+
+    Status Initialize(
+        units::frequency::hertz_t frequency,
+        IsrPointer isr = nullptr, int32_t priority = -1) const override
+    {
+        // NOTE: Same initialization as Timer library's Initialize()
+        system_controller_.PowerUpPeripheral(timer_.channel.power_id);
+        SJ2_ASSERT_FATAL(
+            frequency.to<uint32_t>() != 0,
```
kammce:
```suggestion
            frequency != 0_Hz,
```
You don't need to convert. You can compare the the Hz literal.

library/L1_Peripheral/lpc40xx/capture.hpp
```diff
@@ -0,0 +1,195 @@
...
+    }
+
+    void ConfigureCapture(
+            CaptureChannelNumber channel,
+            CaptureEdgeMode mode) const override
+    {
+        uint32_t shift = (channel == kCaptureChannel1) ? 3 : 0;
+        timer_.channel.timer_register->CCR &= ~(0x03 << shift);
+
+        if (mode)
```
kammce: Every block must have curly braces.

library/L1_Peripheral/lpc40xx/capture.hpp
```diff
@@ -0,0 +1,195 @@
...
+        EnableCaptureInterrupt(kCaptureChannel1, false);
+
+        return Status::kSuccess;
+    }
+
+    void GetCaptureStatus(CaptureChannelNumber channel, CaptureStatus_t &status)
+    {
+        status.flags = timer_.channel.timer_register->IR;
+
+        status.count = (channel == kCaptureChannel0)
```
kammce: Refrain from using ternary.

library/L1_Peripheral/lpc40xx/capture.hpp
```diff
@@ -0,0 +1,195 @@
...
+            ? timer_.channel.timer_register->CR0
+            : timer_.channel.timer_register->CR1;
+    }
+
+    void ConfigureCapture(
+            CaptureChannelNumber channel,
+            CaptureEdgeMode mode) const override
+    {
+        uint32_t shift = (channel == kCaptureChannel1) ? 3 : 0;
+        timer_.channel.timer_register->CCR &= ~(0x03 << shift);
```
kammce: Use `util::bit` libraries for bit manipulation. Do not do bit manipulation directly.

library/L1_Peripheral/lpc40xx/capture.hpp
```diff
@@ -0,0 +1,195 @@
...
+        }
+    }
+ private:
+    const Timer::Channel_t &timer_;
+    const sjsu::SystemController & system_controller_;
+    const sjsu::InterruptController & interrupt_controller_;
+};  // class Capture
+};  // namespace lpc40xx
+};  // namespace sjsu
+
```
kammce: Remove excess newlines. Every file should have 1 and only 1 newline.

library/L1_Peripheral/lpc40xx/test/capture_test.cpp
```diff
@@ -0,0 +1,114 @@
...
+namespace sjsu::lpc40xx
+{
+EMIT_ALL_METHODS(Capture);
+
+TEST_CASE("Testing lpc40xx Capture", "[lpc40xx-capture]")
+{
+    LPC_TIM_TypeDef test_timer_register;
+
+    constexpr int kChannelListSize = 2;
+    constexpr Capture::CaptureChannelNumber kChannelList[kChannelListSize] = {
```
kammce:
```suggestion
    constexpr Capture::CaptureChannelNumber kChannelList[] = {
```
The size constexpr isn't necessary. Remove it.

library/L1_Peripheral/lpc40xx/test/capture_test.cpp
```diff
@@ -0,0 +1,114 @@
...
+        .irq = IRQn::TIMER0_IRQn,
+        .user_callback = &test_timer_isr,
+    };
+
+    Timer::Channel_t test_timer =
+    {
+        .channel = test_timer_partial,
+        .isr = nullptr,
+    };
+
```
kammce: Remove extra line (run `make format-code`)

library/L1_Peripheral/lpc40xx/test/capture_test.cpp
```diff
@@ -0,0 +1,114 @@
...
+        test_subject.Initialize(kTestFrequency, nullptr);
+        int prescaler = (kTestSystemFrequency.to<int32_t>() /
+                kTestPeripheralClockDivider) / kTestFrequency.to<int32_t>();
+        CHECK(test_timer_register.PR == prescaler);
+        CHECK(bit::Read(test_timer_register.TCR, 0) == true);
+    }  // end section initialize
+
+    SECTION("Capture Config")
+    {
+        for (int channel = 0; channel < kChannelListSize; channel++)
```
kammce:
```suggestion
        for (int channel = 0; channel < std::size(kChannelList); channel++)
```
Never use variables/constants in this way. You should always derive the length of an array, variable or object via its label. This way, if you decide to use a different variable for its length, lets say `kChannelListSize2`, your code doesn't break on you, and you don't have to go in and change every place you used `kChannelListSize`.

library/L1_Peripheral/lpc40xx/test/capture_test.cpp
```diff
@@ -0,0 +1,114 @@
...
+{
+    LPC_TIM_TypeDef test_timer_register;
+
+    constexpr int kChannelListSize = 2;
+    constexpr Capture::CaptureChannelNumber kChannelList[kChannelListSize] = {
+        Capture::kCaptureChannel0,
+        Capture::kCaptureChannel1,
+    };
+
+    constexpr int kModeListSize = 4;
```
kammce:
```suggestion
```
Delete as per the comment above.

library/L1_Peripheral/lpc40xx/test/capture_test.cpp
```diff
@@ -0,0 +1,114 @@
...
+
+    SECTION("Capture Initialize")
+    {
+        constexpr units::frequency::hertz_t kTestFrequency = 1_MHz;
+
+        test_subject.Initialize(kTestFrequency, nullptr);
+        int prescaler = (kTestSystemFrequency.to<int32_t>() /
+                kTestPeripheralClockDivider) / kTestFrequency.to<int32_t>();
+        CHECK(test_timer_register.PR == prescaler);
+        CHECK(bit::Read(test_timer_register.TCR, 0) == true);
```
kammce: You should verify that the `mock_interrupt_controller` was called correctly with the correct parameters. This may require that the `test_timer_isr`  be something non-nullptr.

demos/sjtwo/capture/source/main.cpp
```diff
@@ -0,0 +1,85 @@
...
+{
+  constexpr units::frequency::hertz_t kCaptureFrequency = 4_MHz;
+  constexpr units::frequency::hertz_t kGeneratorFrequency = 10_Hz;
+  constexpr float kGeneratorDutyCycle = 0.5f;
+
+  // Configure P1.14 as T2_CAP0
+  LOG_INFO("Configure P1.14 as capture input T2_CAP0");
+  sjsu::lpc40xx::Pin capture_pin(1, 14);
+  capture_pin.SetPinFunction(3);
+  capture_pin.SetAsOpenDrain(false);
```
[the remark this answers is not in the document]
charlesmacd: The capture input signals can be mapped to multiple pins, e.g. T1_CAP0 exists on P1.18 and P3.27 rather than having a 1:1 relationship where a given timer/channel maps to a single pin. One approach is that I could add a parameter to the constructor that takes an enum indicating which pin to use, but if I give a name like 'kPin1_18' then the caller could do something wacky like use that with timer 2 instead of timer 1 for the other parameter. Meaning I'd need some validation on the parameters to look for invalid combinations. Is that the right approach or not quite?

demos/sjtwo/capture/source/main.cpp
```diff
@@ -0,0 +1,85 @@
...
+  capture_unit.ConfigureCapture(sjsu::Capture::kCaptureChannel0,
+    sjsu::Capture::kCaptureEdgeModeBoth);
+  capture_unit.EnableCaptureInterrupt(sjsu::Capture::kCaptureChannel0, true);
+
+  while (1)
+  {
+    sjsu::Delay(1s);
+    float average_count = static_cast<float>(delta_accumulator)
+      / static_cast<float>(num_samples);
+    double frequency = 1.0f / (average_count *
```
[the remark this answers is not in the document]
charlesmacd: To my understanding when you pass a float to a variadic function it gets promoted to double which triggers the double promotion warning, as is the case here using LOG_INFO() with a %f format to print frequency. Changing to double prevents this from happening, but is there a way to use float such that the warning is not triggered?

I was using this as a reference to the issue: https://stackoverflow.com/questions/43011956/why-is-clang-3-8-warning-on-use-of-float-with-f-printf-format-specifier/43012239

library/L1_Peripheral/capture.hpp
charlesmacd: When I run 'make format-code' in the /demos/sjtwo/capture folder it says 'clang-format did not modify any files' and 'make presubmit' indicates the files are OK for the whole source tree. What should I do differently?

library/L1_Peripheral/lpc40xx/capture.hpp
charlesmacd: So restrict the configurations only to the two pins accessible on the SJ2 board? This ties into the other issue of multiple pins mapping to the same capture input depending on the lpc40xx package used (say LQFP80 vs LQFP100). Perhaps rather than having unique timer and channel parameters for the constructor there should be an enum that specifies the timer, channel, and pin -- that way the enum could only contain the valid combinations. But it seems a bit yucky too.

demos/sjtwo/capture/source/main.cpp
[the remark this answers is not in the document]
kammce: The typical implementation is to stick with a single pin for something and allow the user to define their own structure (or copy one) and use their own pin. This way, if someone just wants to use just 1 pin to do capture, they can do this easily. If they want to use a pin we don't currently support, then they have to do the work of setting up the structure.

PR #710, opened 2019-09-12
Add ParallelBus library + ParallelGpio

### LINKED ISSUE #324: Parallel Bus HAL
Drivers like stu7060 build their own structure full of GPIOs to handle manipulating a bus full of gpios. I think it would be fitting to abstract that out of the stu7060 class and make a generic Parallel Bus class. The will allow the easy creation of other customized bus types for other future drivers. 

This abstraction should also benefit of making compiling the stu7060 class test file faster. The parallel test should have fewer test cases, making the compilation time smaller. The stu7060 will simply be using an interface to a parallel bus object which should be lightning fast to compile compared to the large mocked structure for the current test.

kammce:
Resolves #324


projects/hello_world/source/main.cpp
```diff
@@ -2,6 +2,7 @@
 
 #include "utility/log.hpp"
 #include "utility/time.hpp"
+#include "header.hpp"
```
HuangDave: Not sure if this belongs here.

PR #872, opened 2019-09-16
Corrected pins in GPIO Demo, Added SetPullDown function to lpc40 gpio.hpp

LegenDerry9:


library/L1_Peripheral/lpc40xx/gpio.hpp
```diff
@@ -250,6 +250,21 @@ class Gpio final : public sjsu::Gpio
     *interrupt[triggered_port].clear |= (1 << triggered_pin);
   }
 
+  void SetPullDown()
```
kammce: Why is this here? You don't need to put in an arbitrary method into gpio to set the pull up. Setting pull up modes is a `Pin` specific thing, not a GPIO specific thing. Use the `GetPin()` method in the gpio library to set it as pull up.

PR #880, opened 2019-10-02
Corrected pins in GPIO Demo

LegenDerry9:


demos/sjtwo/eeprom/makefile
```diff
@@ -1,5 +1,15 @@
...
+
+ifndef SJSU_DEV2_BASE
+$(info +-------------- SJSU-Dev2 Location file not found --------------+)
+$(info |                                                               |)
+$(info |        Run ./setup from within the SJSU-Dev2's folder         |)
+$(info |                                                               |)
+$(info +---------------------------------------------------------------+)
+$(error )
+endif
+
```
kammce: Not sure why this was not there originally, but it is good that you added this here.

PR #883, opened 2019-10-03
Converted Accelerometer Driver from google drive document to markdown.

hsorathia:


hsorathia:
I'm very confused as to what I just did.

PR #891, opened 2019-10-06
Add L1 PulseCapture

### LINKED ISSUE #505: LPC40xx: Pulse Capture

charlesmacd:
Resolve #505

### DIFF
--- a/demos/sjtwo/pulse_capture/source/main.cpp
+++ b/demos/sjtwo/pulse_capture/source/main.cpp
@@ -1,0 +1,69 @@
+#include <cstdint>
+
+#include "L1_Peripheral/lpc40xx/pulse_capture.hpp"
+#include "L1_Peripheral/lpc40xx/pwm.hpp"
+
+volatile uint32_t previous_count;
+volatile uint32_t current_count;
+volatile uint32_t delta_accumulator;
+volatile uint32_t num_samples;
+
+/* Sample our PWM signal generator output at 4 MHz */
+constexpr units::frequency::hertz_t kCaptureFrequency = 4_MHz;
+
+/* Generate a 10 Hz square wave using PWM as a test signal generator */
+constexpr units::frequency::hertz_t kGeneratorFrequency = 10_Hz;
+
+/* Configure the test signal duty cycle to 50% */
+constexpr float kGeneratorDutyCycle = 0.5f;
+
+sjsu::lpc40xx::PulseCapture capture_unit(
+    sjsu::lpc40xx::PulseCapture::Channel::kCaptureTimer2,
+    sjsu::lpc40xx::PulseCapture::CaptureChannelNumber::kChannel0,
+    kCaptureFrequency);
+
+void CaptureISR(sjsu::PulseCapture::CaptureStatus_t status)
+{
+  previous_count = current_count;
+  current_count  = status.count;
+  delta_accumulator += (current_count - previous_count);
+  ++num_samples;
+}
+
+int main()
+{
+  // Configure P1.14 as T2_CAP0
+  LOG_INFO("Configure P1.14 as capture input T2_CAP0");
+
+  // Have capture unit sample T2_CAP0 at 1 MHz
+  capture_unit.Initialize(CaptureISR);
+
+  // Create 60 Hz square wave on P2.0, that we can capture
+  LOG_INFO("Creating PWM output on P2.0");
+  sjsu::lpc40xx::Pwm signal_generator(sjsu::lpc40xx::Pwm::Channel::kPwm0);
+  signal_generator.Initialize(kGeneratorFrequency);
+  signal_generator.SetDutyCycle(kGeneratorDutyCycle);
+
+  // Enable capture interrupt for both edges
+  LOG_INFO("Enabling capture");
+  capture_unit.ConfigureCapture(sjsu::PulseCapture::CaptureEdgeMode::kBoth);
+  capture_unit.EnableCaptureInterrupt(true);
+
+  while (1)
+  {
+    // Wait long enough for the capture interrupt to trigger at least once
+    sjsu::Delay(1s);
+
+    float average_count =
+        static_cast<float>(delta_accumulator) / static_cast<float>(num_samples);
+    float frequency =
+        1.0f / (average_count * (1.0f / kCaptureFrequency.to<float>())) / 2.0f;
+    LOG_INFO("Input frequency = %.3f Hz, Expected = %d",
+             static_cast<double>(frequency),
+             kGeneratorFrequency.to<int>());
+    delta_accumulator = 0;
+    num_samples       = 0;
+  }
+
+  return 0;
+}
--- a/library/L1_Peripheral/lpc40xx/pulse_capture.hpp
+++ b/library/L1_Peripheral/lpc40xx/pulse_capture.hpp
@@ -1,0 +1,324 @@
+#pragma once
+
+#include <cstdint>
+
+#include "L0_Platform/lpc40xx/LPC40xx.h"
+#include "L1_Peripheral/interrupt.hpp"
+#include "L1_Peripheral/lpc40xx/pin.hpp"
+#include "L1_Peripheral/lpc40xx/system_controller.hpp"
+#include "L1_Peripheral/lpc40xx/timer.hpp"
+#include "L1_Peripheral/lpc40xx/pulse_capture.hpp"
+#include "L1_Peripheral/timer.hpp"
+#include "L1_Peripheral/pulse_capture.hpp"
+#include "utility/bit.hpp"
+#include "utility/log.hpp"
+#include "utility/status.hpp"
+#include "utility/units.hpp"
+
+namespace sjsu
+{
+namespace lpc40xx
+{
+/// Pulse capture library
+class PulseCapture final : public sjsu::PulseCapture
+{
+ public:
+  /// List of available timers to capture from
+  enum CaptureTimerNumber : uint8_t
+  {
+    kTimer0,
+    kTimer1,
+    kTimer2,
+    kTimer3
+  };
+
+  /// List of available channels per timer to capture from
+  enum CaptureChannelNumber : uint8_t
+  {
+    kChannel0,
+    kChannel1
+  };
+
+  /// Structure used to define attributes of a timer capture channel
+
+  struct CaptureChannelPartial_t
+  {
+    /// Pointer to memory-mapped timer peripheral registres
+    LPC_TIM_TypeDef * timer_register;
+    /// Peripheral ID of the timer peripheral to power on at initialization
+    sjsu::SystemController::PeripheralID power_id;
+    /// Interrupt number associated with this timer
+    IRQn irq;
+    /// Callback invoked during a capture event
+    CaptureIsr * user_callback;
+    /// Pin corresponding to timer capture channel #0
+    const sjsu::Pin & capture_pin0;
+    /// Pin corresponding to timer capture channel #1
+    const sjsu::Pin & capture_pin1;
+    /// Pointer to variable containing capture channel number
+    CaptureChannelNumber * channel_number;
+  };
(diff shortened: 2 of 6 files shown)


library/L1_Peripheral/lpc40xx/capture.hpp
```diff
@@ -0,0 +1,285 @@
...
+        units::frequency::hertz_t frequency,
+        IsrPointer isr = nullptr, int32_t priority = -1) const override
+    {
+        // NOTE: Same initialization as Timer library's Initialize()
+        system_controller_.PowerUpPeripheral(timer_.channel.power_id);
+        SJ2_ASSERT_FATAL(
+            frequency != 0_Hz,
+            "Cannot have zero ticks per microsecond, please choose 1 or more.");
+
+        // Set Prescale register for Prescale Counter to milliseconds
```
kammce: Isn't this just creating a divider. I'm not sure what you mean by milliseconds.

library/L1_Peripheral/lpc40xx/capture.hpp
```diff
@@ -0,0 +1,285 @@
...
+                .interrupt_service_routine = timer_.isr,
+                .enable_interrupt = true,
+                .priority = priority
+        });
+
+        // Disable both capture channels
+        ConfigureCapture(kCaptureChannel0, kCaptureEdgeModeNone);
+        ConfigureCapture(kCaptureChannel1, kCaptureEdgeModeNone);
+        EnableCaptureInterrupt(kCaptureChannel0, false);
+        EnableCaptureInterrupt(kCaptureChannel1, false);
```
kammce: Should you do this first before you register the ISR. If these are defined as true before you get to this point, there is a chance that an event can occur between the registration of the ISR and this disabling of the ISR. This is more evident in a threaded system where the task controlling this peripheral is of low priority and not getting as much CPU time.

library/L1_Peripheral/lpc40xx/capture.hpp
```diff
@@ -0,0 +1,285 @@
...
+        else
+        {
+            timer_.channel.capture_pin0.SetPinFunction(3);
+        }
+
+        static constexpr bit::Mask kModeBits[2] =
+        {
+            bit::CreateMaskFromRange(0, 1),
+            bit::CreateMaskFromRange(3, 4)
+        };
```
kammce: Something is also doable with bit::Masks is that they are actually just a structure with a uint8_t for position and width. Rather than making an array of these you could simply do this:

```suggestion
        static constexpr bit::Mask kModeBits =
        {
            .position =  3 * which,
            .width = 2,
        };
```
This will eliminate some of the code below. Be sure to comment why this is being done though.

library/L1_Peripheral/lpc40xx/capture.hpp
```diff
@@ -0,0 +1,285 @@
...
+
+        channel.timer_register->IR = bit::Insert(channel.timer_register->IR,
+            kResetCaptureInterrupts, kCaptureInterruptFlagBit);
+    }
+
+    struct CaptureStatus_t
+    {
+        uint32_t count;
+        uint32_t flags;
+    };
```
kammce: Move this to the top with the other structures.

library/L1_Peripheral/lpc40xx/capture.hpp
```diff
@@ -0,0 +1,285 @@
...
+        {
+            bit::CreateMaskFromRange(2),
+            bit::CreateMaskFromRange(5)
+        };
+
+        int which = 0;
+        if (channel == kCaptureChannel1)
+        {
+            which = 1;
+        }
```
kammce: If you set kCaptureChannel1 ==> 1 and kCaptureChannel0 == 0, then you can do this and omit the comparison.

```suggestion
        which = static_cast<int>(capture_channel_)
```

library/L1_Peripheral/lpc40xx/test/capture_test.cpp
```diff
@@ -0,0 +1,153 @@
...
+void DummyCallback(void) {}
+
+TEST_CASE("Testing lpc40xx Capture", "[lpc40xx-capture]")
+{
+    LPC_TIM_TypeDef test_timer_register;
+
+    constexpr Capture::CaptureChannelNumber kChannelList[] = {
+      Capture::kCaptureChannel0,
+      Capture::kCaptureChannel1,
+    };
```
kammce: make sure that `make format-code` hits this file as well.

library/L1_Peripheral/lpc40xx/test/capture_test.cpp
```diff
@@ -0,0 +1,153 @@
...
+        .capture_pin1 = capture1_input_pin
+    };
+
+    const Capture::CaptureChannel_t kTestTimer =
+    {
+        .channel = kTestTimerPartial,
+        .isr = sjsu::lpc40xx::Capture::TimerHandler<kTestTimerPartial>
+    };
+
+    memset(&test_timer_register, 0, sizeof(test_timer_register));
```
kammce: Can you put this right below the creation of the `test_timer_register`. Makes it easier to see how the registers was configured before it was assigned to something. For example, its referenced in the kTestPartial, then cleared, which may give someone the idea, that if the tests are failing its due to the fact that you did not clear your register data, when you actually did, but they missed it because they read the code top to bottom, then found what they believed to be the actual cause of the problem, when in actuality, this part is fine, and something they did broke the code.

Basically, could result in a red herring.

library/L1_Peripheral/lpc40xx/test/capture_test.cpp
```diff
@@ -0,0 +1,153 @@
...
+#include "utility/enum.hpp"
+#include "utility/status.hpp"
+#include "utility/units.hpp"
+
+
+namespace sjsu::lpc40xx
+{
+EMIT_ALL_METHODS(Capture);
+
+void DummyCallback(void) {}
```
kammce: Could you either put this in a anonymous namespace so the link name does not class against other tests in the `sjsu::lpc40xx` (think about how many other tests probably have a function for an ISR named DummyCallback). The other option is to create a non-capturing lambda in the TEST_CASE.
```suggestion
namespace {
void DummyCallback(void) {}
}
```

library/L1_Peripheral/lpc40xx/test/capture_test.cpp
```diff
@@ -0,0 +1,153 @@
...
+          position = 3;
+        }
+
+        for (size_t mode = 0; mode < std::size(kModeList); mode++)
+        {
+          test_subject.ConfigureCapture(kChannelList[channel], kModeList[mode]);
+          CHECK(bit::Extract(test_timer_register.CCR, position, 2) ==
+                kModeList[mode]);
+        }
+        }
```
kammce: Incorrect indentation for the curly brace. Please use `make format-code` to fix this.

library/L1_Peripheral/lpc40xx/test/capture_test.cpp
```diff
@@ -0,0 +1,153 @@
...
+
+    Capture test_subject(kTestTimer,
+        mock_system_controller.get(),
+        mock_interrupt_controller.get());
+
+    SECTION("Capture Initialize")
+    {
+        constexpr units::frequency::hertz_t kTestFrequency = 1_MHz;
+
+        test_subject.Initialize(kTestFrequency, &DummyCallback);
```
kammce: Missing a test for interrupt.RegisterIsr().

library/L1_Peripheral/lpc40xx/test/capture_test.cpp
```diff
@@ -0,0 +1,153 @@
...
+        test_subject.EnableCaptureInterrupt(kChannelList[channel], false);
+        CHECK(bit::Read(test_timer_register.CCR, position) == false);
+        }
+    }  // end section capture interrupt
+
+    SECTION("Capture Interrupt Handler")
+    {
+      test_timer_register.IR = 0;
+      kTestTimer.isr();
+      CHECK(bit::Extract(test_timer_register.IR, 4, 2) == 0b11);
```
kammce: Missing a test for a call to DummyCallback, make sure that your ISR does in fact call it. It may be helpful to use the FFF library to mock the calllback and get invocation history.

library/L1_Peripheral/lpc40xx/test/capture_test.cpp
```diff
@@ -0,0 +1,153 @@
...
+    }  // end section capture interrupt handler
+}  // end test case
+}  // namespace sjsu::lpc40xx
+
+
```
kammce: Too many newline characters here.

library/L1_Peripheral/lpc40xx/test/capture_test.cpp
```diff
@@ -0,0 +1,153 @@
+#include <cstdint>
+
+#include "L1_Peripheral/cortex/interrupt.hpp"
+#include "L1_Peripheral/interrupt.hpp"
+#include "L0_Platform/lpc40xx/LPC40xx.h"
+#include "L1_Peripheral/lpc40xx/timer.hpp"
+#include "L1_Peripheral/lpc40xx/capture.hpp"
+#include "L4_Testing/testing_frameworks.hpp"
+#include "utility/enum.hpp"
```
kammce: I don't think you used this at all in this file.

library/L1_Peripheral/lpc40xx/test/capture_test.cpp
```diff
@@ -0,0 +1,153 @@
+#include <cstdint>
+
+#include "L1_Peripheral/cortex/interrupt.hpp"
+#include "L1_Peripheral/interrupt.hpp"
+#include "L0_Platform/lpc40xx/LPC40xx.h"
```
kammce: Put this at the top as 0 is higher in the alphabetical order.

demos/sjtwo/pulse_capture/source/main.cpp
```diff
@@ -0,0 +1,70 @@
+#include <cstdint>
+
+#include "L1_Peripheral/lpc40xx/timer.hpp"
```
kammce: I don't think you need this here.

library/L1_Peripheral/lpc40xx/pulse_capture.hpp
```diff
@@ -0,0 +1,334 @@
...
+
+  /// List of available channels per timer to capture from
+  enum CaptureChannelNumber : uint8_t
+  {
+    kChannel0,
+    kChannel1
+  };
+
+  /// Structure used to define attributes of a timer capture channel
+
```
kammce:
```suggestion
```
Delete extra space between structure and comment.

library/L1_Peripheral/lpc40xx/pulse_capture.hpp
```diff
@@ -0,0 +1,334 @@
...
+        {
+          status.count = channel.timer_register->CR0;
+        }
+        else
+        {
+          status.count = channel.timer_register->CR1;
+        }
+
+        (*channel.user_callback)(status);
+        }
```
kammce:
```suggestion
        }
```
Got a bit confused here and thought you were not clearing the interrupt flags here. Can you move this curly brace into the correct place.

library/L1_Peripheral/pulse_capture.hpp
```diff
@@ -0,0 +1,45 @@
...
+                            int32_t interrupt_priority = -1) const = 0;
+
+  /// Select edge type to capture on
+  virtual void ConfigureCapture(CaptureEdgeMode mode) const = 0;
+
+  /// Enable or disable capture interrupts
+  virtual void EnableCaptureInterrupt(bool enabled) const = 0;
+};
+
+} /* namespace sjsu */
```
kammce:
```suggestion
}  // namespace sjsu::lpc40xx
```
Can you change this to the following suggestion like the rest of your code.

PR #925, opened 2019-10-21
Adding Infrared Receiver Interface + Tsop752 Design Document

HuangDave:


documentation/design_docs/L2/tsop752.md
```diff
@@ -0,0 +1,131 @@
...
+# Detailed Design
+## API
+```C++
+namespace sjsu
+{
+class Tsop752 final : public InfraredReceiver
+{
+ public:
+  constexpr explicit Tsop752(sjsu::Capture & capture,
+                             sjsu::Timer & timer);
```
kammce: I don't understand why you need both a timer and a capture peripheral.

documentation/design_docs/L2/tsop752.md
```diff
@@ -0,0 +1,131 @@
...
+N/A
+
+# Testing Plan
+## Unit Testing Scheme
+The `Capture` and `Timer` modules shall be mocked and stubbed to provide test
+inputs for the `Tsop752` driver.
+
+|       Function      |    Expected Result    |
+|:--------------------|:----------------------|
+|Initialize           | Should return the failure `Status` `Initialize()` of either `Capture` or `Timer` are stubbed to not return `Status::kSuccess`.<br/> Should return `Status::kSuccess` when `Initialize()` of both `Capture` and `Timer` are stubbed to return `Status::kSuccess`. |
```
kammce: Can you fix this? This looks really messed up. Try and keep it within 80 cols.

documentation/design_docs/L2/tsop752.md
HuangDave: The timer helps automatically detect the end of a transmission after not receiving a pulse after a duration or should the user just specify how many pulses to be expected per frame?

documentation/design_docs/L2/tsop752.md
HuangDave: Starting a new line would cause the sentence to appear in a new cell. Should I just use bullet points instead and stay away from tables?

kammce: Yeah, use bullet points instead.

kammce:
I'm merging this in now, but I feel that there is a way to remove the dependency on the timer class. I believe that using Uptime() should be sufficient.

PR #927, opened 2019-10-23
Updated EEPROM driver to use bit utility

harmandersihra:
This commit updates all the bit manipulation in the eeprom lpc40xx driver to use the bit utility.

### DIFF
--- a/library/L1_Peripheral/lpc40xx/eeprom.hpp
+++ b/library/L1_Peripheral/lpc40xx/eeprom.hpp
@@ -6,6 +6,8 @@
 #include "L0_Platform/lpc40xx/LPC40xx.h"
 #include "L1_Peripheral/lpc40xx/system_controller.hpp"
 
+#include "utility/bit.hpp"
+
 namespace sjsu
 {
 namespace lpc40xx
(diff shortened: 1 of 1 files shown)


library/L1_Peripheral/lpc40xx/eeprom.hpp
```diff
@@ -94,18 +99,19 @@ class Eeprom final : public sjsu::Eeprom
   void Write(const uint8_t * wdata, uint32_t full_address,
              size_t count) const override
   {
-    constexpr uint8_t kMax6Bits = 0b11'1111;
+    bit::Mask k_lower6_bits = bit::CreateMaskFromRange(0, 5);
```
kammce:
```suggestion
    constexpr bit::Mask kLower6Bits = bit::CreateMaskFromRange(0, 5);
```
Do the same for the others.

library/L1_Peripheral/lpc40xx/eeprom.hpp
```diff
@@ -94,18 +99,19 @@ class Eeprom final : public sjsu::Eeprom
   void Write(const uint8_t * wdata, uint32_t full_address,
              size_t count) const override
   {
-    constexpr uint8_t kMax6Bits = 0b11'1111;
+    bit::Mask k_lower6_bits = bit::CreateMaskFromRange(0, 5);
+    bit::Mask k_upper6_bits = bit::CreateMaskFromRange(6, 11);
```
kammce: Do no proceed a variable name with a k if it is not a constant. This should be a constant and should be Cap Camel case with a lowercase k in front.

PR #957, opened 2019-11-25
Migrate from IsrPointer to std::function

### LINKED ISSUE #952: Migrate from IsrPointer to std::function for non-ISRs
IsrPointer is required for functions on the vector table, but they are utilized in other places as well, as a way to add a callback for various functions. This is extremely limiting in that it requires that a stateless function pointer must be utilized. Usage of cstdlib std::function opens the gates to utilizing capturing lambdas, std::bind for class member function execution, usage of functors, and much, much more.

There is a compile time cost, in that the gcc-arm-eabi-7 header file for <functional> takes a while to parse as well as a space cost where each std::function occupies 32 bytes of space, thus an array of such functions will take up 4x the amount of space compared to the 4 bytes for a function pointer.

kammce:
Split usage of IsrPointer into two different categories of interrupt
callables.

For actual interrupts that get put into the IVT, such as an interrupt
service routine, they will be named `InterruptHandler`.
For calling a callable objects from within an interrupt service routine,
these will be called `InterruptCallback`.

`InterruptHanlder` is defined as a function pointer of void(*)(void),
where as `InterruptCallback` a std::function<void(void)>, which allows
the usage of function pointers as well as capturing lambdas, bindings
to class methods, and usage of functors.

Caveats to this is that the <functional> header increases compile time
and occupies 32 bytes of space per object, which is 4x the size of a
pointer.

Resolves #952

### DIFF
--- a/demos/arm_cortex/system_timer/source/main.cpp
+++ b/demos/arm_cortex/system_timer/source/main.cpp
@@ -15,7 +15,7 @@ int main()
   sjsu::cortex::SystemTimer system_timer(
       sjsu::GetInactive<sjsu::SystemController>());
 
-  system_timer.SetInterrupt(DemoSystemIsr);
+  system_timer.SetCallback(DemoSystemIsr);
   system_timer.SetTickFrequency(10_Hz);
   system_timer.StartTimer();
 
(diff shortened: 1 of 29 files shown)


kammce:
Merged vis #961

Issue #597, opened 2019-07-11
Timer: Add Start(), Stop() and Reset() to interface


kammce:
Without Start(), Stop() and Reset(), attempting to time things out properly can be very difficult, especially when the count frequency is high and the match registers are small.

Issue #534, opened 2019-06-12
IR Receiver: TSOP752


kammce:

ntwong0:
Please assign me to this

PR #971, opened 2019-11-30
Adding Tsop752 Infrared Receiver

HuangDave:
- Added InfraredReceiver interface.
- Added example project for the Tsop752 using the SJOne.
- Added Start(), Stop(), and Reset() to Timer.
- Changed Initialize() in lpc40xx::Timer to not automatically enable the timer
  peripheral during initialization. This allows more flexibility in controlling
  when the timer can be started.
- Updated Interface Receiver Interface and Tsop752 design docs.

Resolves: #534
Resolves: #597

### DIFF
--- a/demos/sjone/ir_receiver/source/main.cpp
+++ b/demos/sjone/ir_receiver/source/main.cpp
@@ -1,0 +1,82 @@
+#include "L1_Peripheral/lpc17xx/system_controller.hpp"
+#include "L1_Peripheral/lpc17xx/pulse_capture.hpp"
+#include "L1_Peripheral/lpc17xx/timer.hpp"
+#include "L2_HAL/communication/tsop752.hpp"
+#include "utility/infrared_algorithms.hpp"
+#include "utility/log.hpp"
+
+enum RemoteCode : uint16_t
+{
+  kPower           = 0x0A5F,
+  kVolumeUp        = 0x32CD,
+  kVolumeDown      = 0xB24D,
+  kToggleBluetooth = 0x40BF,
+};
+/// @see https://www.sbprojects.net/knowledge/ir/nec.php
+constexpr sjsu::infrared::PulseDurationConfiguration_t kNecConfiguration = {
+  .header_mark_duration  = 9000us,
+  .header_space_duration = 4500us,
+  .data_duration         = 560us,
+  .logic_high_duration   = 1690us,
+  .logic_low_duration    = 560us,
+  .encoding_type         = sjsu::infrared::PulseDurationType::kDistance,
+  .tolerance             = 0.25f,  // value may vary based on device latency
+  .uses_repeat_frames    = true,
+  .header_repeat_space   = 2250us,
+};
+
+void HandleDataFrame(const sjsu::infrared::DataFrame_t * data_frame)
+{
+  // Attempt to decode the received frame using the NEC protocol.
+  const sjsu::infrared::DecodedFrame_t kDecodedFrame =
+      sjsu::infrared::Decode(data_frame, kNecConfiguration);
+  // Do nothing if the frame is invalid or the command is from another remote
+  if (!kDecodedFrame.is_valid)
+  {
+    return;
+  }
+  // Check whether a repeat frame was decoded. A repeat frame after the initial
+  // frame is continuously sent while a button is held down.
+  if (kDecodedFrame.is_repeat)
+  {
+    LOG_INFO("Button is held down repeating previous action.");
+  }
+  else
+  {
+    constexpr auto kCommandMask = sjsu::bit::CreateMaskFromRange(0, 15);
+    const uint16_t kCommand     = static_cast<uint16_t>(
+        sjsu::bit::Extract(kDecodedFrame.data, kCommandMask));
+    switch (kCommand)
+    {
+      case RemoteCode::kPower: LOG_INFO("Power Button Pressed"); break;
+      case RemoteCode::kVolumeUp: LOG_INFO("Increasing Volume"); break;
+      case RemoteCode::kVolumeDown: LOG_INFO("Decreasing Volume"); break;
+      case RemoteCode::kToggleBluetooth:
+        LOG_INFO("Toggling Bluetooth Mode");
+        break;
+      default: LOG_INFO("Button Pressed"); break;
+    }
+  }
+}
+
+int main()
+{
+  LOG_INFO("Starting Tsop752 IR Receiver Example...");
+  // Using timeout of 9ms since the header mark of the NEC remote is 9ms long.
+  constexpr std::chrono::microseconds kTimeout               = 9ms;
+  constexpr units::frequency::hertz_t kPulseCaptureFrequency = 1_MHz;
+
+  sjsu::lpc17xx::PulseCapture pulse_capture(
+      sjsu::lpc17xx::PulseCaptureChannel::kCaptureTimer1,
+      sjsu::lpc17xx::PulseCapture::CaptureChannelNumber::kChannel0,
+      kPulseCaptureFrequency);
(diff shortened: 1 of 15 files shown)


library/L2_HAL/communication/infrared_receiver.hpp
```diff
@@ -0,0 +1,37 @@
...
+/// infrared (IR) data.
+class InfraredReceiver
+{
+ public:
+  /// IR data frame containing the timestamps for the received de-modulated IR
+  /// signal.
+  struct DataFrame_t
+  {
+    /// Maximum size of the pulse_buffer.
+    inline static constexpr uint32_t kMaxPulseBufferSize = 100;
```
kammce:
```suggestion
    inline static constexpr uint32_t kMaxPulseBufferSize = 100;
```
The size for this is quite large. Does it need to be this large?

kammce: Are the timestamps between the pulses so large that we need 32 bits of resolution or can 8bit or 16bit suffice?

library/L2_HAL/communication/test/tsop752_test.cpp
```diff
@@ -0,0 +1,143 @@
...
+  Mock<Timer> mock_timer;
+
+  Fake(Method(mock_pulse_capture, ConfigureCapture),
+       Method(mock_pulse_capture, EnableCaptureInterrupt));
+  Fake(Method(mock_timer, SetMatchBehavior),
+       Method(mock_timer, Start),
+       Method(mock_timer, Stop),
+       Method(mock_timer, Reset));
+
+  const Status kFailureStatus[] = {
```
kammce:
```suggestion
  const std::array kFailureStatus[] = {
```
Prefer to use std::array vs c-style arrays.

library/L2_HAL/communication/test/tsop752_test.cpp
```diff
@@ -0,0 +1,143 @@
...
+
+  Tsop752 ir_receiver(mock_pulse_capture.get(), mock_timer.get(), kTimeout);
+
+  SECTION("Initialize")
+  {
+    SECTION("PulseCapture peripheral initialization failure")
+    {
+      mock_timer.ClearInvocationHistory();
+      When(Method(mock_timer, Initialize)).AlwaysReturn(Status::kSuccess);
+      for (uint8_t i = 0; i < std::size(kFailureStatus); i++)
```
kammce:
```suggestion
      for (size_t i = 0; i < kFailureStatus.size(); i++)
```
Use the size_t for the rest of the for loops here.

library/L2_HAL/communication/test/tsop752_test.cpp
```diff
@@ -0,0 +1,143 @@
...
+      {
+        const Status kExpectedStatus = kFailureStatus[i];
```
kammce: Use the https://github.com/catchorg/Catch2/blob/master/docs/logging.md#top Logging macro as a way to disambiguate between errors. This is important as a failed assertion here will show the line number and the case that was checked, but will not include which index the failure occured on.

library/L2_HAL/communication/test/tsop752_test.cpp
```diff
@@ -0,0 +1,143 @@
...
+    }
+    SECTION("Handle end of frame")
+    {
+      bool user_callback_invoked = false;
+      InfraredReceiver::DataReceivedHandler mock_user_callback =
+          [&](auto data_frame) {
+            user_callback_invoked = true;
+            for (uint8_t i = 0; i < kExpectedPulseEdgeCount; i++)
+            {
+              CHECK(data_frame->pulse_buffer[i] == kTestTimestamps[i]);
```
kammce: A really good place to put a logging macro.

demos/sjone/ir_receiver/source/main.cpp
```diff
@@ -0,0 +1,146 @@
...
+    return;
+  }
+  constexpr auto kCommandMask = sjsu::bit::CreateMaskFromRange(0, 15);
+  const uint16_t kCommand     = static_cast<uint16_t>(
+      sjsu::bit::Extract(kDecodedFrame.data, kCommandMask));
+  switch (kCommand)
+  {
+    case kPower: LOG_INFO("Power Button Pressed"); break;
+    case kVolumeUp: LOG_INFO("Increasing Volume"); break;
+    case kVolumeDown: LOG_INFO("Decreasing Volume"); break;
```
kammce: Add a default case as well.

demos/sjone/ir_receiver/source/main.cpp
```diff
@@ -0,0 +1,146 @@
...
+/// Example:
+///    _____________________                       ___________
+///   |                     |                     |
+///   ^                     v                     ^
+///   |                     |_____________________|
+///
+///   |<------ mark ------->|<------ space ------>|
+///
+/// @see: http://lirc.sourceforge.net/remotes/samsung/BN59-00678A
+const DecodedFrame_t DecodeSamsungRemote(
```
kammce: This may be good to put this algorithm somewhere in the libraries.

HuangDave: Would it be better to put this in the InfraredReceiver class as a helper function or to put it in a utility/infrared.hpp? This function can be modified a bit to be a generic decoding algorithm for pulse width encoding.

kammce: I'm honestly not sure where the best place would be. I think it would make the most sense in the InfraredReceiver header. Not sure if having it within the class makes the most sense. Could just be an inline function. It could also be in another file named "infrared_algorithms.h". 

Making it generic would be great!

documentation/design_docs/L2/infrared_receiver_interface.md
```diff
@@ -36,13 +55,13 @@ class InfraredReceiver
   struct DataFrame_t
   {
     inline static constexpr uint32_t kMaxPulseBufferSize = 100;
-    uint32_t pulse_buffer[kMaxPulseBufferSize];
-    uint32_t number_of_pulses;
+    volatile uint32_t pulse_buffer[kMaxPulseBufferSize];
```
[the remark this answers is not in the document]
kammce: This has not be answered yet.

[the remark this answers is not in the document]
HuangDave: For this I was thinking the values might not be interrupt safe or not thread safe when using RTOS.

library/L1_Peripheral/lpc40xx/timer.hpp
```diff
@@ -130,6 +135,26 @@ class Timer final : public sjsu::Timer
     return timer_.peripheral->TC;
   }
 
+  void Start() const override
+  {
+    timer_.peripheral->TCR = bit::Set(
+        timer_.peripheral->TCR, TimerControlRegister::kEnableBit.position);
```
kammce:
```suggestion
        timer_.peripheral->TCR, TimerControlRegister::kEnableBit);
```
I added a version of this that does not require you to pull out the position field.

library/L2_HAL/communication/test/tsop752_test.cpp
```diff
@@ -0,0 +1,154 @@
...
+                 .Using(kExpectedTimerMatchCount,
+                        sjsu::Timer::MatchAction::kInterrupt,
+                        kExpectedTimerMatchRegister))
+          .Once();
+    }
+  }
+
+  SECTION("Interrupt Callbacks")
+  {
+    constexpr std::array<uint32_t, 6> kTestTimestamps = {
```
kammce:
```suggestion
    constexpr std::array<uint32_t> kTestTimestamps = {
```
C++17 added in template deduction guides for std::array that allows you to omit the array length and allow the initializer_list<> to define it.

PR #980, opened 2019-12-11
Add Units Polarity Checking and Lambdas to LTC4150

### LINKED ISSUE #990: Add Units, Polarity Checking and Lambda ISRs to LTC4150
I got this!
#980

evanugarte:
Now the `LTC4150` class is able to check if the battery is charging by using the user specified polarity pin. All ISR handlers are also made into lambdas, and units are used for different values such as charge, resistance and voltage.

Resolves #990


library/L2_HAL/sensors/battery/ltc4150.hpp
```diff
@@ -4,95 +4,74 @@
...
+    pol_pin_.SetAsInput();
     int_pin_.SetAsInput();
-    int_pin_.AttachInterrupt(isr_handler_, Gpio::Edge::kEdgeFalling);
+    pol_pin_.AttachInterrupt(
+        [this]() {
+          polarity_ =
+              (pol_pin_.Read()) ? Polarity::kDischarging : Polarity::kCharging;
+        },
+        Gpio::Edge::kEdgeFalling);
+    int_pin_.AttachInterrupt([this]() { pulses_++; }, Gpio::Edge::kEdgeFalling);
```
kammce:
```suggestion
    int_pin_.AttachInterrupt([this]() { 
    if (polarity == Polarity::kDischarging)
    {
      pulses_--; 
     } 
     else 
     {
        pulses_++;
     }
    }, Gpio::Edge::kEdgeFalling);
```
The polarity can change at any time during usage of a coulomb counter thus, you need to check it to decide if you should decrement the pulses or increment.

library/L2_HAL/sensors/battery/ltc4150.hpp
```diff
@@ -4,95 +4,74 @@
...
-  /// @return the calculated mAh from the TickHandler instance
-  float GetBatterymAh() const override
+  /// @return the calculated mAh
+  units::charge::milliampere_hour_t GetCharge() const override
   {
-    return get_mah_();
+    /// 3600 is X 32.55 is kGvF and 0.00005 is kRsense
+    auto sign = (polarity_ == Polarity::kCharging)
+                    ? units::dimensionless::scalar_t(-1)
+                    : units::dimensionless::scalar_t(1);
```
kammce: See comment about about subtracting and adding to pulses based on the polarity. This doesn't make sense here as the pin could change at anytime during usage. If it was unchanging, we should could simply have the developer give us  a  bool parameter in the constructor to figure out which state it will be in rather than using a Gpio.

library/L2_HAL/sensors/battery/ltc4150.hpp
```diff
@@ -4,95 +4,74 @@
...
-  /// Gpio pin to check polarity output from the LTC4150.
-  sjsu::Gpio & pol_;
-  /// function pointer to grab current battery mAh
-  float (*get_mah_)(void) = nullptr;
-  /// function pointer to isr handler
-  void (*isr_handler_)(void) = nullptr;
+  sjsu::Gpio & pol_pin_;
+  units::impedance::ohm_t resistance_;
+  int pulses_;
+  Polarity polarity_;
```
kammce:
```suggestion
    std::atomic<Polarity> polarity_;
```
Use atomic here to make it clear that a set, read or get operation on this field will occur atomically, since ISR have the capability to nest themselves.

library/L2_HAL/sensors/battery/ltc4150.hpp
```diff
@@ -4,95 +4,74 @@
...
   sjsu::Gpio & int_pin_;
-  /// Gpio pin to check polarity output from the LTC4150.
-  sjsu::Gpio & pol_;
-  /// function pointer to grab current battery mAh
-  float (*get_mah_)(void) = nullptr;
-  /// function pointer to isr handler
-  void (*isr_handler_)(void) = nullptr;
+  sjsu::Gpio & pol_pin_;
+  units::impedance::ohm_t resistance_;
+  int pulses_;
```
kammce:
```suggestion
  std::atomic<int32_t> pulses_;
```
Use atomic here to make it clear that a set, read or get operation on this field will occur atomically. since ISRs have the capability to nest themselves.

library/L2_HAL/sensors/battery/test/ltc4150_test.cpp
```diff
@@ -1,38 +1,60 @@
...
 #include "L4_Testing/testing_frameworks.hpp"
+#include "L1_Peripheral/interrupt.hpp"
+#include "L1_Peripheral/gpio.hpp"
+#include "utility/units.hpp"
 
 namespace sjsu
 {
 #define APPROX_EQUALITY(expected, actual, resolution) \
   CHECK(-resolution < (actual - expected));           \
   CHECK((actual - expected) < resolution);
```
kammce: Rather than using this macro, you can use the Catch2's `Approx()` matcher instead.

library/L2_HAL/sensors/battery/test/ltc4150_test.cpp
```diff
@@ -1,38 +1,60 @@
...
 
+auto GetLambda(sjsu::InterruptCallback & isr)
+{
+  return [&isr](sjsu::InterruptCallback callback,
+                [[maybe_unused]] sjsu::Gpio::Edge edge) { isr = callback; };
```
kammce:
```suggestion
                sjsu::Gpio::Edge) { isr = callback; };
```
If you simply do not give the argument a name, it will automatically be ignored, which means you can remove the `[[maybe_unused]]` attribute.

library/L2_HAL/sensors/battery/ltc4150.hpp
evanugarte: I added an initial state parameter that was of type Ltc4150::Polarity instead of bool. Would this work? I also got rid of this line since the sign is accounted for in the updated polarity ISR.

projects/hello_world/project.mk
```diff
@@ -1,3 +1 @@
-USER_TESTS += $(LIBRARY_DIR)/L1_Peripheral/cortex/test/dwt_counter_test.cpp
-USER_TESTS += $(LIBRARY_DIR)/L1_Peripheral/cortex/test/interrupt_test.cpp
-USER_TESTS += $(LIBRARY_DIR)/L1_Peripheral/test/interrupt_test.cpp
+USER_TESTS += $(LIBRARY_DIR)/L2_HAL/sensors/battery/test/ltc4150_test.cpp
```
kammce: I think you did this for testing. Can you revert to previous tests

library/L2_HAL/sensors/battery/ltc4150.hpp
```diff
@@ -4,95 +4,86 @@
...
   {
   }
 
   /// Initialize hardware, setting pins as inputs and attaching ISR handlers to
   /// the interrupt pin.
-  void Initialize() const override
+  void Initialize() override
   {
-    pol_.SetAsInput();
+    pol_pin_.SetAsInput();
```
kammce: One thing that should be done at initialization is that the pin should be read in order to figure out what the initial polarity is, rather than waiting for the polarity to change. In most applications, the polarity will be fixed from the point of power on.

library/L2_HAL/sensors/battery/test/ltc4150_test.cpp
```diff
@@ -46,28 +65,37 @@ TEST_CASE("Test LTC4150 Coulomb Counter/ Battery Gas Gauge", "[ltc4150]")
...
-    backup_isr.IsrHandler();
-    backup_isr.IsrHandler();
-    APPROX_EQUALITY(backup_counter.GetBatterymAh(), 1.02407f, kResolution);
+    primary_counter.Initialize();
+    primary_input_isr();
+    primary_input_isr();
+    primary_input_isr();
+    primary_input_isr();
+    primary_pol_isr();
+    CHECK(primary_counter.GetCharge().to<float>() == Approx(-0.068271f));
```
kammce: Can you derive the approximate values you use here so that they do not look like mere magic numbers. Same with the stuff below.

library/L2_HAL/sensors/battery/test/ltc4150_test.cpp
```diff
@@ -46,28 +65,37 @@ TEST_CASE("Test LTC4150 Coulomb Counter/ Battery Gas Gauge", "[ltc4150]")
...
-    backup_isr.IsrHandler();
-    backup_isr.IsrHandler();
-    backup_isr.IsrHandler();
-    APPROX_EQUALITY(backup_counter.GetBatterymAh(), 1.02407f, kResolution);
+    primary_counter.Initialize();
+    primary_input_isr();
+    primary_input_isr();
+    primary_input_isr();
+    primary_input_isr();
+    primary_pol_isr();
```
kammce: Calling primary_pol_isr() does nothing here to change the output and thus does not test anything. Add another test section that utitilizes the pol_isr. 

It would be good to have one where you set the default direction as discharge, run x5 pulses, then change the polarity via the pol_isr, then run 5x pulses. This should result in a net 0 charge in total.

Another test would be to have a non symmetric number of charge/discharge pulses and show that the polarity change does in fact work.

projects/hello_world/project.mk
evanugarte: Done. Sorry, i forgot to revert it

library/L2_HAL/sensors/battery/ltc4150.hpp
evanugarte: Done.

library/L2_HAL/sensors/battery/test/ltc4150_test.cpp
evanugarte: Done.

library/L2_HAL/sensors/battery/test/ltc4150_test.cpp
evanugarte: Done.

library/L2_HAL/sensors/battery/test/ltc4150_test.cpp
evanugarte: All the uses of `Approx()` work great, except for the last test case. I scheduled a meeting with you today to show the problem I am encountering,

evanugarte: Done! Thanks for showing me [epsilon](https://github.com/catchorg/Catch2/blob/master/docs/assertions.md#floating-point-comparisons)!!

library/L2_HAL/sensors/battery/ltc4150.hpp
evanugarte: I removed the supplied parameter, and I read the gpio at `Initialize()` [here](https://github.com/kammce/SJSU-Dev2/blob/8b224131585cd1212a5a0c085fdf5660e2f82f8c/library/L2_HAL/sensors/battery/ltc4150.hpp#L61). To incorporate the `Initialize()` read in my tests, I added a change in mock behavor, for example with the cancel and change in sign test cases [here](https://github.com/kammce/SJSU-Dev2/blob/8b224131585cd1212a5a0c085fdf5660e2f82f8c/library/L2_HAL/sensors/battery/test/ltc4150_test.cpp#L155)

PR #1014, opened 2019-12-23
Resolve issue with LPC17xx FreeRTOS failing

kammce:
Clean up parts of code that are not used in startup.

### DIFF
--- a/library/L0_Platform/lpc17xx/startup.cpp
+++ b/library/L0_Platform/lpc17xx/startup.cpp
@@ -68,14 +68,6 @@ extern "C"
 
   void vPortSetupTimerInterrupt(void)  // NOLINT
   {
-    interrupt_controller.Enable({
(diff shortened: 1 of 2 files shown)


library/L0_Platform/lpc17xx/startup.cpp
```diff
@@ -68,14 +68,6 @@ extern "C"
...
   void vPortSetupTimerInterrupt(void)  // NOLINT
   {
-    interrupt_controller.Enable({
-        .interrupt_request_number = sjsu::cortex::SVCall_IRQn,
-        .interrupt_handler        = vPortSVCHandler,
-    });
-    interrupt_controller.Enable({
-        .interrupt_request_number = sjsu::cortex::PendSV_IRQn,
-        .interrupt_handler        = xPortPendSVHandler,
-    });
```
kammce: This was not needed if we are injecting the function pointers directly into the IVT.

PR #1020, opened 2019-12-27
Break in master due to temperature.hpp rename

### LINKED ISSUE #1019: Break in master due to temperature.hpp renamed to temperature_sensor.hpp
Causes the `find-missing-tests` target to fail because the file was renamed but the flag in the L2.mk file was not changed.

kammce:
Resolves #1019


kammce:
@HuangDave commenting on this to show this update. I really need to document all of these flags and what they mean.

PR #1090, opened 2020-03-19
Fixing newlib in FreeRTOS

HuangDave:
See: http://www.nadler.com/embedded/NXP_newlibAndFreeRTOS.html


kammce:
Rather than enabling reentrant newlib, we may want to move to using FreeRTOS's heap allocators and --wrap all of the BS malloc/free implementations that newlib uses. They cause the binary to bloat up and its kind of a pain. We currently have the issue because we use std::function<> for interrupts which pulls in dynamic memory allocation.

PR #1104, opened 2020-03-23
Refactoring St7066u to use ParallelBus

HuangDave:


library/L2_HAL/displays/lcd/st7066u.hpp
```diff
@@ -96,69 +90,51 @@ class St7066u
...
+    kEnablePin.Set(sjsu::Gpio::kHigh);
+    kDataBus.Initialize();
+    kDataBus.SetDirection(sjsu::Gpio::Direction::kOutput);
 
-    WriteCommand(static_cast<uint8_t>(Command::kDefaultDisplayConfiguration) |
-                 static_cast<uint8_t>(kBusMode) |
-                 static_cast<uint8_t>(kDisplayMode) |
-                 static_cast<uint8_t>(kFontStyle));
+    WriteCommand(Value(Command::kDefaultDisplayConfiguration) |
+                 Value(kBusMode) | Value(kDisplayMode) | Value(kFontStyle));
```
kammce:
```suggestion
    WriteCommand(Value(Command::kDefaultDisplayConfiguration) |
                 Value(kBusMode) | Value(kDisplayMode) | Value(kFontStyle));
```

See `enum.hpp` and look for the `SJ2_ENABLE_BITMASK_OPERATORS` so that you can OR these together without having to use Value so often. Also

If you use that, then you won't even need to use the `Value()` method for the enums becaue you can have `WriteCommand` take a composed `Command` enum object.

library/L2_HAL/displays/lcd/st7066u.hpp
```diff
@@ -167,56 +143,46 @@ class St7066u
...
-    kControlPins.d7.Set(sjsu::Gpio::State((nibble >> 3) & 0x01));
-    kControlPins.d6.Set(sjsu::Gpio::State((nibble >> 2) & 0x01));
-    kControlPins.d5.Set(sjsu::Gpio::State((nibble >> 1) & 0x01));
-    kControlPins.d4.Set(sjsu::Gpio::State((nibble >> 0) & 0x01));
+    kEnablePin.Set(sjsu::Gpio::State::kHigh);
+    kRegisterSelectPin.Set(sjsu::Gpio::State(operation));
+    kReadWritePin.Set(sjsu::Gpio::State::kLow);
+
+    kDataBus.Write(nibble);
     sjsu::Delay(1ms);
```
kammce: Does this have to be 1ms? or can it be shorter? Seems like a long time. Also, Delay can be used with nanoseconds precision now.

library/L2_HAL/displays/lcd/st7066u.hpp
```diff
@@ -167,56 +143,46 @@ class St7066u
...
   {
-    kControlPins.e.Set(sjsu::Gpio::State::kHigh);
-    kControlPins.rs.Set(sjsu::Gpio::State(operation));
-    kControlPins.rw.Set(sjsu::Gpio::State::kLow);
-    // set nibble on 4-bit data bus
-    kControlPins.d7.Set(sjsu::Gpio::State((nibble >> 3) & 0x01));
-    kControlPins.d6.Set(sjsu::Gpio::State((nibble >> 2) & 0x01));
-    kControlPins.d5.Set(sjsu::Gpio::State((nibble >> 1) & 0x01));
-    kControlPins.d4.Set(sjsu::Gpio::State((nibble >> 0) & 0x01));
+    kEnablePin.Set(sjsu::Gpio::State::kHigh);
```
kammce:
```suggestion
    kEnablePin.SetHigh();
```
Just an FYI, may be a bit more concise if you simply use the helper functions for this.

library/L2_HAL/displays/lcd/st7066u.hpp
```diff
@@ -167,56 +143,46 @@ class St7066u
...
+    kEnablePin.Set(sjsu::Gpio::State::kLow);
+    kEnablePin.Set(sjsu::Gpio::State::kHigh);
   }
 
   /// Transfers a command or data byte to the device.
   ///
   /// @param operation Operation transfer type.
   /// @param byte      Byte to transfer.
-  void WriteByte(WriteOperation operation, uint8_t byte)
+  void WriteByte(WriteOperation operation, uint8_t byte) const
```
kammce: Is WriteByte or WriteNibble actually any different? They look basically the same now that we are using ParallelBus. You might be able to consolidate them together and simply check if you need to run it twice for a nibble in the places where WriteByte/Nibble is used. 

If this works, then you can reduce the code size of the object.

library/L2_HAL/displays/lcd/st7066u.hpp
```diff
@@ -247,28 +213,28 @@ class St7066u
 
   /// Clears all characters on the display by sending the clear display command
   /// to the device.
-  virtual void ClearDisplay()
+  void ClearDisplay() const
   {
     WriteCommand(Command::kClearDisplay);
     sjsu::Delay(2ms);  // Clear display operation requires 1.52ms
```
kammce:
```suggestion
    sjsu::Delay(1.6ms);  // Clear display operation requires 1.52ms
```
You can do that now. But if you want to be safe, keeping it 2ms is also fine.

library/L2_HAL/displays/lcd/test/st7066u_test.cpp
```diff
@@ -0,0 +1,343 @@
...
+
+    SECTION("With 4-bit Bus")
+    {
+      bus_mode                         = St7066u::BusMode::kFourBit;
+      const uint8_t kConfigurationData = Value(kDefaultConfiguration) |
+                                         Value(bus_mode) | Value(kDisplayMode) |
+                                         Value(kFontStyle);
+      // For 4-bit bus mode, the most significant nibble is sent first followed
+      // by the least significant nibble.
+      expected_data = {
```
kammce: A little confusing. Maybe put a comment stating that this array gets checked in the `AlwaysDo` statement above.

library/L2_HAL/displays/lcd/test/st7066u_test.cpp
```diff
@@ -0,0 +1,343 @@
...
+  SECTION("Write Command using 4-bit Bus")
+  {
+    constexpr Gpio::State kExpectedRegisterSelect =
+        Gpio::State(St7066u::WriteOperation::kCommand);
+    uint8_t expected_data;
+    St7066u lcd = kFourBitBusLcd;
+
+    SECTION("ClearDisplay")
+    {
+      expected_data = static_cast<uint8_t>(St7066u::Command::kClearDisplay);
```
kammce:
```suggestion
      expected_data = Value(St7066u::Command::kClearDisplay);
```
You can use `Value` here.

library/L2_HAL/displays/lcd/st7066u.hpp
HuangDave: Unfortunately this doesn't seem to work for multiple different enums.

kammce: Ah that makes sense. If they are not the same enum class then my eable bitmask operators does not work.

HuangDave:
Can't pull in changes since CI is broken @kammce .

PR #1106, opened 2020-03-25
Add SD library refactor

### LINKED ISSUE #348: SD Card Driver: Simplify Function Implementation with Memory Block Class
At its current state, the Sd Card Driver is not terribly easy to use with arrays of blocks. It imposes a lot of extra work on the user to make sure their block constructs are of the correct block size. This could be simplified with the addition of a designated MemoryBlock class that can be used to represent individual blocks of SD Card Memory.

### LINKED ISSUE #349: SD Card Driver: Split Sd::Mount() into separate, single-purpose functions
The `Sd::Mount()` function has to perform many operations to mount the SD card and prime it for CRUD operations. These jobs could instead be divided into their own functions to increase the class's modularity and make it easier to debug.

kammce:
- Upgraded 3P printf to support a greater range of
- Upgraded printf config options in support of the latest printf library

Resolves #349
Resolves #348


kammce:
Error is from a change in XCode server and not from the code provided.

PR #1111, opened 2020-03-29
Adding Unit Tests for Timer

### LINKED ISSUE #423: Add Timer class unit test

HuangDave:
Resolves #423

### DIFF
--- a/library/L1_Peripheral/lpc40xx/test/timer_test.cpp
+++ b/library/L1_Peripheral/lpc40xx/test/timer_test.cpp
@@ -4,5 +4,132 @@
 namespace sjsu::lpc40xx
 {
 EMIT_ALL_METHODS(Timer);
-TEST_CASE("Testing lpc40xx Timer", "[lpc40xx-timer]") {}
+
+TEST_CASE("Testing lpc40xx Timer", "[lpc40xx-timer]")
+{
+  LPC_TIM_TypeDef local_timer_registers;
+  memset(&local_timer_registers, 0, sizeof(local_timer_registers));
+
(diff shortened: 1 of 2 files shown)


library/L1_Peripheral/lpc40xx/test/timer_test.cpp
```diff
@@ -4,5 +4,132 @@
...
+  }
+
+  SECTION("GetCount")
+  {
+    CHECK(local_timer_registers.TC == 0);
+    CHECK(timer.GetCount() == 0);
+
+    constexpr uint32_t kExpectedCount = 12'345;
+    local_timer_registers.TC          = kExpectedCount;
+    CHECK(timer.GetCount() == kExpectedCount);
```
kammce:
```suggestion
    // Setup
    constexpr uint32_t kExpectedCount = 12'345;
    local_timer_registers.TC          = kExpectedCount;
    
    // Exercise
    uint32_t actual_count = timer.GetCount();
    
    // Verify
    CHECK(actual_count == kExpectedCount);
```

Can you format your tests with a "setup", "exercise" and "verify" regions with comments. This is something that I learned about at work. And it really helps to organize your test cases.

I would also make another test sections just for testing that GetCount() defaults to 0 when TC is zero, rather than overload it in this test section. Each test section should have 1 setup, exercise, verify. But you can loop the exercise and verify as long as the thing you are testing does not change subjects.

kammce: Make sure to do this for the other sections as well.

kammce:
CI is broken for one of the OSX images. I'll go and fix that. I think you should be able to merge this yourself without me. But if you cannot ping me when the changes are made and I can force merge it in.

PR #1113, opened 2020-03-29
CI: Removed Coveralls install in xcode images

### LINKED ISSUE #1112: XCode11.3 Breaking CI due to pip install failure
https://travis-ci.com/github/kammce/SJSU-Dev2/builds/155253125

kammce:
This must have been residual code left over when Coveralls was called on
all platforms. But in this case, it is only called in Linux, so there is
no need to call it on any xcode images.

Resolves #1112

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -35,15 +35,11 @@ matrix:
               branch: master
         - os: osx
           osx_image: xcode10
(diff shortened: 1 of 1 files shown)


kammce:
@HuangDave FYI

PR #1116, opened 2020-03-31
Refactor Task class

### LINKED ISSUE #1058: C++ Tasks shouldn't need to have a scheduler passed to them.
The semantics of doing the following:

```
Scheduler scheduler;
Task task(..., &scheduler);
```

Does not seem as natural as:

```
Scheduler scheduler;
Task task(...);
scheduler.Add(task);
```

HuangDave:
Resolves #1058


HuangDave:
Will need to resolve #1090 before the example projects can run on hardware as printing will cause hard fault.

PR #1129, opened 2020-04-04
Fixing FreeRTOS example projects

HuangDave:
Increased the task's stack size from 512 bytes to 1024 bytes to allow
enough static memory for printf. Previously, the stack size of 512 did
not provide enough memory which caused a hard fault when printf is
invoked after the FreeRTOS scheduler is started.


kammce:
Can you update the commit message and this PR to reflect the reason why we need to increase the stack size of the tasks due to the stack memory based printf library.

PR #1131, opened 2020-04-05
Adding MSP432P401R Pin and Gpio class

### LINKED ISSUE #869: MSP432P401R: GPIO

HuangDave:
Resolves #869

### DIFF
--- a/demos/msp432p401r/blinker/source/main.cpp
+++ b/demos/msp432p401r/blinker/source/main.cpp
@@ -1,24 +1,22 @@
 #include <cstdio>
 
 #include "L0_Platform/msp432p401r/msp432p401r.h"
-#include "utility/bit.hpp"
+#include "L1_Peripheral/msp432p401r/gpio.hpp"
 #include "utility/log.hpp"
 #include "utility/time.hpp"
 
 int main()
 {
   sjsu::LogInfo("Starting MSP432P401R LED Blinker Demo...");
+
   // Configure the on-board P1.0 LED to be initially turned on.
-  constexpr sjsu::bit::Mask kLedBitMask = sjsu::bit::CreateMaskFromRange(0);
-  sjsu::msp432p401r::P1->DIR            = sjsu::bit::Set(
-      sjsu::msp432p401r::P1->DIR, sjsu::bit::CreateMaskFromRange(0));
-  sjsu::msp432p401r::P1->OUT =
-      sjsu::bit::Set(sjsu::msp432p401r::P1->OUT, kLedBitMask);
+  sjsu::msp432p401r::Gpio p1_0(1, 0);
+  p1_0.SetAsOutput();
+  p1_0.SetHigh();
 
   while (true)
   {
-    sjsu::msp432p401r::P1->OUT =
-        sjsu::bit::Toggle(sjsu::msp432p401r::P1->OUT, kLedBitMask);
+    p1_0.Toggle();
     sjsu::Delay(1s);
   }
   return 0;
--- a/demos/msp432p401r/system_controller/project.mk
+++ b/demos/msp432p401r/system_controller/project.mk
@@ -1,3 +1,4 @@
 USER_TESTS += $(LIBRARY_DIR)/L1_Peripheral/msp432p401r/test/system_controller_test.cpp
+
 PLATFORM = msp432p401r
 JTAG = xds110
--- a/demos/msp432p401r/system_controller/source/main.cpp
+++ b/demos/msp432p401r/system_controller/source/main.cpp
@@ -2,8 +2,9 @@
 
 #include "L0_Platform/msp432p401r/msp432p401r.h"
 #include "L1_Peripheral/cortex/system_timer.hpp"
+#include "L1_Peripheral/msp432p401r/gpio.hpp"
+#include "L1_Peripheral/msp432p401r/pin.hpp"
 #include "L1_Peripheral/msp432p401r/system_controller.hpp"
-#include "utility/bit.hpp"
 #include "utility/log.hpp"
 #include "utility/time.hpp"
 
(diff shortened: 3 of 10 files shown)


demos/msp432p401r/system_controller/project.mk
```diff
@@ -1,3 +1,6 @@
 USER_TESTS += $(LIBRARY_DIR)/L1_Peripheral/msp432p401r/test/system_controller_test.cpp
+USER_TESTS += $(LIBRARY_DIR)/L1_Peripheral/msp432p401r/test/pin_test.cpp
+USER_TESTS += $(LIBRARY_DIR)/L1_Peripheral/msp432p401r/test/gpio_test.cpp
```
kammce, line 3: Move these into the blinker demo's project.mk

library/L1_Peripheral/msp432p401r/test/gpio_test.cpp
```diff
@@ -0,0 +1,250 @@
...
+        .registers =
+            reinterpret_cast<DIO_PORT_Interruptable_Type &>(local_dio_j),
+    },
+  };
+
+  SECTION("SetDirection")
+  {
+    for (size_t i = 0; i < test_pins.size(); i++)
+    {
+      Gpio & pin                = test_pins[i].gpio;
```
kammce, line 103: Write a comment 

```
// Setup
```
Here.

library/L1_Peripheral/msp432p401r/test/pin_test.cpp
```diff
@@ -0,0 +1,243 @@
...
+      const uint8_t kPortNumber = pin.GetPort();
+      const uint8_t kPinNumber  = pin.GetPin();
+
+      INFO("port: " << static_cast<size_t>(kPortNumber));
+      INFO("pin: " << static_cast<size_t>(kPinNumber));
+
+      for (uint8_t function_code = 0; function_code <= 0b111; function_code++)
+      {
+        // Setup
+        INFO("function: 0b" << std::bitset<3>(function_code));
```
kammce, line 119: Oh that is neat! Gotta remember that for later.

library/L1_Peripheral/msp432p401r/pin.hpp
```diff
@@ -0,0 +1,182 @@
...
+    }
+    else
+    {
+      *drive_strength_register = bit::Clear(*drive_strength_register, pin_);
+    }
+  }
+
+  void SetAsOpenDrain(bool) const override
+  {
+    // TODO(#): Should use Error-v2.
```
kammce, line 121:
```suggestion
    // TODO(#1134): Should use Error-v2.
```

library/L1_Peripheral/msp432p401r/pin.hpp
```diff
@@ -0,0 +1,182 @@
...
+    SJ2_ASSERT_FATAL(false, "This function is not implemented");
+  }
+
+  void SetAsAnalogMode(bool) const override
+  {
+    // TODO(#): Should use Error-v2.
```
kammce, line 127:
```suggestion
    // TODO(#1134): Should use Error-v2.
```

library/L1_Peripheral/msp432p401r/pin.hpp
```diff
@@ -0,0 +1,182 @@
...
+    SJ2_ASSERT_FATAL(false, "This function is not implemented");
```
kammce, line 128: Is this really a fatal error? Does the MSP432p401r not have ADCs? If the MCU does have an ADC, but need not set any particular bits for Analog control, this it makes sense to simply have this function do nothing, as this is already handled by setting the pin function.

In the case above with open drain, it makes sense to emit and error or fatally halt because proceeding can be dangerous to the hardware. But in this case, if the system continues to work even without this implemented and a HAL abstraction calls this, it shouldn't Halt().

demos/msp432p401r/blinker/source/main.cpp
```diff
@@ -1,24 +1,23 @@
 #include <cstdio>
 
 #include "L0_Platform/msp432p401r/msp432p401r.h"
+#include "L1_Peripheral/msp432p401r/gpio.hpp"
 #include "utility/bit.hpp"
```
kammce, line 5: It looks like you are not using this include anymore.

demos/msp432p401r/system_controller/source/main.cpp
```diff
@@ -2,6 +2,8 @@
 
 #include "L0_Platform/msp432p401r/msp432p401r.h"
 #include "L1_Peripheral/cortex/system_timer.hpp"
+#include "L1_Peripheral/msp432p401r/gpio.hpp"
+#include "L1_Peripheral/msp432p401r/pin.hpp"
 #include "L1_Peripheral/msp432p401r/system_controller.hpp"
 #include "utility/bit.hpp"
```
kammce, line 8: It looks like you are not using this include anymore.

library/L1_Peripheral/msp432p401r/pin.hpp
```diff
@@ -0,0 +1,188 @@
...
+    }
+  }
+
+  void SetAsOpenDrain(bool) const override
+  {
+    // TODO(#1134): Should use Error-v2.
+    SJ2_ASSERT_FATAL(false, "This function is not implemented");
+  }
+
+  void SetAsAnalogMode(bool) const override {}
```
kammce, line 129: Add a comment here stating that SetAnalogMode is not required here in order to enable ADCs.

demos/msp432p401r/system_controller/source/main.cpp
HuangDave: Pin is used to configure the clock out pins and GPIO is used for the LED.

PR #1135, opened 2020-04-07
Migrate Pin to Error-v2

### LINKED ISSUE #1134: Error 2.0: Migrate Pin to Returns<T>

HuangDave:
Resolves #1134

### DIFF
--- a/library/L1_Peripheral/pin.hpp
+++ b/library/L1_Peripheral/pin.hpp
@@ -2,6 +2,8 @@
 
 #include <cstdint>
 
+#include "utility/status.hpp"
+
 namespace sjsu
 {
 /// Abstraction interface for physical electrical "pins".
@@ -26,9 +28,10 @@ namespace sjsu
 class Pin
 {
  public:
-  // ==============================
+  // ===========================================================================
   // Interface Defintions
-  // ==============================
+  // ===========================================================================
+
   /// Defines the set of internal resistance connections to a pin. Read specific
   /// enumeration constant comments/documentation to understand more about what
   /// each one does.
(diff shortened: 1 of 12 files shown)


kammce:
This looks great, but I fear that making this change to integral parts of the code base that people may be using before summer is a bad idea. I made this change to storage because it was a new interface made during the academic school year. For now, I can review these and we can discuss them in the PR, but this MUST NOT be merged into the main code base.

library/L1_Peripheral/pin.hpp
```diff
@@ -93,26 +98,29 @@ class Pin
   /// the job of the L1 peripheral that uses this pin to manage its own function
   /// code usage.
   ///
-  /// @param function - pin function code
+  /// @param function pin function code
   virtual void SetPinFunction(uint8_t function) const = 0;
```
kammce, line 102: I feel like this could also return an error, do you agree?

library/L1_Peripheral/pin.hpp
```diff
@@ -93,26 +98,29 @@ class Pin
...
+
   /// Set pin's resistor pull, setting ot either no resistor pull, pull down,
   /// pull up and repeater.
   ///
   /// @param resistor - which resistor setup you would like.
   virtual void SetPull(Resistor resistor) const = 0;
```
kammce, line 108: Couldn't this return an error if the resistor state desired does not exist for that platform?

library/L1_Peripheral/pin.hpp
HuangDave: I agree, we could check for whether the function code is valid or not.

kammce:
Closing this for now.

PR #1143, opened 2020-04-16
Refactor MSP432 System Controller

### LINKED ISSUE #1137: Migrate msp432xx to SystemController 2.0

HuangDave:
Resolves #1137

### DIFF
--- a/library/L1_Peripheral/msp432p401r/system_controller.hpp
+++ b/library/L1_Peripheral/msp432p401r/system_controller.hpp
@@ -18,6 +18,25 @@ namespace msp432p401r
 class SystemController final : public sjsu::SystemController
 {
  public:
+  /// Contains the device id for each of the available clock system modules.
+  class Modules
+  {
+   public:
+    //! @cond Doxygen_Suppress
+    static constexpr auto kAuxiliaryClock       = PeripheralID::Define<0>();
+    static constexpr auto kMasterClock          = PeripheralID::Define<1>();
+    static constexpr auto kSubsystemMasterClock = PeripheralID::Define<2>();
+    static constexpr auto kLowSpeedSubsystemMasterClock =
+        PeripheralID::Define<3>();
+    static constexpr auto kBackupClock           = PeripheralID::Define<4>();
+    static constexpr auto kLowFrequencyClock     = PeripheralID::Define<5>();
+    static constexpr auto kVeryLowFrequencyClock = PeripheralID::Define<6>();
+    static constexpr auto kReferenceClock        = PeripheralID::Define<7>();
+    static constexpr auto kModuleClock           = PeripheralID::Define<8>();
+    static constexpr auto kSystemClock           = PeripheralID::Define<9>();
+    //! @endcond
+  };
+
   /// The available internal oscillators for the clock system module.
   enum class Oscillator : uint8_t
   {
@@ -43,7 +62,7 @@ class SystemController final : public sjsu::SystemController
   /// kAuxiliary, kMaster, kSubsystemMaster, kLowSpeedSubsystemMaster, and
   /// kBackup are the primary clock signals.
   ///
-  /// @see Figure 6-1 in the MSP432P4xx Reference Manual.
+  /// @see Figure 6-1. Clock System Block Diagram
   ///      https://www.ti.com/lit/ug/slau356i/slau356i.pdf#page=380
   enum class Clock : uint8_t
   {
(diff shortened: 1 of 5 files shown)


library/L1_Peripheral/msp432p401r/system_controller.hpp
```diff
@@ -72,11 +72,36 @@ class SystemController final : public sjsu::SystemController
...
+    kDivideBy4   = 0b010,
+    kDivideBy8   = 0b011,
+    kDivideBy16  = 0b100,
+    kDivideBy32  = 0b101,
+    kDivideBy64  = 0b110,
+    kDivideBy128 = 0b111,
+    //! @endcond
+  };
+
+  class Peripherals
```
kammce, line 90: I was certain I was going to hate the name Peripherals for this and now I see why. What you did is the correct thing, but now we need to come up with a better semantic word for what these things are. Do you have any ideas?

HuangDave:
Initial changes for MSP432 system controller. Just want to get it reviewed/checked before updating tests.

library/L1_Peripheral/msp432p401r/system_controller.hpp
```diff
@@ -145,111 +226,158 @@ class SystemController final : public sjsu::SystemController
...
 
-    WaitForClockReadyStatus(Clock::kMaster);
-    SetClockSource(Clock::kMaster, Oscillator::kDigitallyControlled);
-    SetClockSource(Clock::kSubsystemMaster, Oscillator::kDigitallyControlled);
+    switch (clock_configuration_.backup.clock_source)
+    {
+      case Oscillator::kLowFrequency:
+      case Oscillator::kReference:
+      default: break;
+    }
```
kammce, line 315: Whats the purpose of this switch case here?

library/L1_Peripheral/msp432p401r/system_controller.hpp
```diff
@@ -145,111 +226,158 @@ class SystemController final : public sjsu::SystemController
...
-  ///
-  /// @param peripheral_divider Divider value to set. Only the following
-  ///                           dividers are available: 1, 2, 4, 8, 16, 32, 64,
-  ///                           128.
-  void SetPeripheralClockDivider(PeripheralID,
-                                 uint8_t peripheral_divider) const
-  {
-    SetClockDivider(Clock::kSubsystemMaster, peripheral_divider);
+    auxiliary_clock_rate_ =
+        aclk / kDividers[Value(clock_configuration_.auxiliary.divider)];
```
kammce, line 322: I think this would more optimal rather than using a look up table. It may be interesting to test this.
 
```suggestion
    auxiliary_clock_rate_ =
        aclk / (1 << Value(clock_configuration_.auxiliary.divider));
```

Lookup table option:
- Permant 4 x 8 bytes (32 bytes) 
- Array lookup

Shift left:
- Potentially a single 16-bit thumb instruction to perform the `1 << amount` arithmetic.

library/L1_Peripheral/msp432p401r/system_controller.hpp
```diff
@@ -72,11 +72,36 @@ class SystemController final : public sjsu::SystemController
     kSystem,
   };
 
+  /// The available clock dividers for the primary clocks.
+  enum ClockDivider : uint8_t
```
kammce, line 76:
```suggestion
  enum class ClockDivider : uint8_t
```

library/L1_Peripheral/msp432p401r/system_controller.hpp
```diff
@@ -376,91 +514,156 @@ class SystemController final : public sjsu::SystemController
...
+        1500_kHz, 3_MHz, 6_MHz, 12_MHz, 24_MHz, 48_MHz
+      };
+      const auto kCenterFrequency = kDcoCenterFrequencies[dco_frequency_select];
+      const float kFrequencyDifference =
+          (kTargetFrequency - kCenterFrequency).to<float>();
+      const float kCalibration =
+          (1.0f + dco_constant * static_cast<float>(768 - dco_calibration));
+      const int16_t kTuneValue =
+          static_cast<int16_t>((kFrequencyDifference * kCalibration) /
+                               (kTargetFrequency.to<float>() * dco_constant));
```
kammce, line 607: Do this in multiple steps to make this a bit cleaner/easier to understand. Also know that doing division between things of the same unit returns integers as well.

library/L1_Peripheral/msp432p401r/system_controller.hpp
```diff
@@ -376,91 +514,156 @@ class SystemController final : public sjsu::SystemController
...
+      // =======================================================================
+      UnlockClockSystemRegisters();
+      {
+        clock_system->CTL0 = bit::Insert(clock_system->CTL0, kTuneValue,
+                                         Control0Register::kTuningSelect);
+        clock_system->CTL0 =
+            bit::Insert(clock_system->CTL0, dco_frequency_select,
+                        Control0Register::kFrequencySelect);
+        clock_system->CTL0 =
+            bit::Set(clock_system->CTL0, Control0Register::kEnable);
```
kammce, line 620: It would be better to read CTL0 into a variable, and perform the insertion operations on the variable to allow for easier optimization by the compiler. Unless these steps must be done in this specific order, in which case, add a comment to contextualize that.

library/L1_Peripheral/msp432p401r/system_controller.hpp
HuangDave: What about naming it as Modules?

library/L1_Peripheral/msp432p401r/system_controller.hpp
HuangDave: Completely missed this, meant to set `bclk` here.

HuangDave:
Update unit tests.

library/L1_Peripheral/msp432p401r/system_controller.hpp
kammce: Yeah, you can rename it to that. Need to figure out a good name for the PeripheralID variable.

library/L1_Peripheral/msp432p401r/system_controller.hpp
```diff
@@ -128,14 +191,91 @@ class SystemController final : public sjsu::SystemController
...
+      /// On reset, the digitally controlled clock is used to drive the master
+      /// clock and the subsystem master clocks. This value should be set to
+      /// false if an alternate clock source is used to drive those clocks.
+      bool enabled = true;
+      /// The target DCO output frequency. On reset, DCO outputs 3 MHz.
+      units::frequency::hertz_t frequency = 3_MHz;
+    };
+
+    //! @cond Doxygen_Suppress
+    AuxiliaryClock_t auxiliary              = {};
```
kammce, line 272:
```suggestion
    struct {
      /// @note Only the reference clock, very low frequency clock, or the low
      ///       frequency clock can be used as the clock source.
      Oscillator clock_source = Oscillator::kReference;
      /// Clock divider for the auxiliary clock.
      ClockDivider divider = ClockDivider::kDivideBy1;
    } auxiliary              = {};
```

This is something I did for lpc40xx. What do you think? It reduced the amount of code quite a bit. I feel like creating these structures, since they have absolutely no use outside of this class is not necessary. Thus we can create an anonymous structure inline here.

PR #1203, opened 2020-05-08
Adding utility methods GetMaximumValue and Voltage to ADC Interface

### LINKED ISSUE #1011: ADC utility method MaximumValue() and Voltage()
```
uint32_t GetMaximumValue()
{
    return (1 << adc_.GetActiveBits()) - 1);
}
```
Useful utility to get what the ratio between max adc value.

`Voltage()`
Would return the voltage based on a voltage reference.

HuangDave:
Resolves #1011

### DIFF
--- a/library/L1_Peripheral/lpc40xx/adc.hpp
+++ b/library/L1_Peripheral/lpc40xx/adc.hpp
@@ -217,9 +217,15 @@ class Adc final : public sjsu::Adc
     return bit::Read(adc_base->CR, Control::kBurstEnable);
   }
 
-  /// @param channel: Passed channel descriptor object. See Channel_t and
+  /// @param channel Passed channel descriptor object. See Channel_t and
   ///        Channel documentation for more details about how to use this.
-  explicit constexpr Adc(const Channel_t & channel) : channel_(channel) {}
+  /// @param reference_voltage The ADC reference voltage in microvolts.
+  explicit constexpr Adc(const Channel_t & channel,
+                         units::voltage::microvolt_t reference_voltage = 3.3_V)
+      : channel_(channel), kReferenceVoltage(reference_voltage)
+  {
+  }
+
   Status Initialize() const override
   {
     auto & system     = sjsu::SystemController::GetPlatformController();
@@ -268,6 +274,11 @@ class Adc final : public sjsu::Adc
     return kActiveBits;
   }
 
+  units::voltage::microvolt_t ReferenceVoltage() const override
+  {
+    return kReferenceVoltage;
+  }
+
  private:
   bool HasConversionFinished() const
   {
@@ -296,6 +307,7 @@ class Adc final : public sjsu::Adc
   }
 
   const Channel_t & channel_;
+  const units::voltage::microvolt_t kReferenceVoltage;
 };
 }  // namespace lpc40xx
 }  // namespace sjsu
--- a/library/L1_Peripheral/test/adc_test.cpp
+++ b/library/L1_Peripheral/test/adc_test.cpp
@@ -1,0 +1,95 @@
+#include "L1_Peripheral/adc.hpp"
+#include "L4_Testing/testing_frameworks.hpp"
+
+namespace sjsu
+{
+TEST_CASE("Testing ADC Interface", "[adc-interface]")
+{
+  Mock<Adc> mock_adc;
+
+  Fake(Method(mock_adc, Initialize));
+  Fake(Method(mock_adc, Read));
+  Fake(Method(mock_adc, GetActiveBits));
+  Fake(Method(mock_adc, ReferenceVoltage));
+
+  Adc & test_subject = mock_adc.get();
+
+  SECTION("GetMaximumValue")
+  {
+    // Setup
+    uint32_t expected_active_bits;
+
+    SECTION("8-bit Resolution")
+    {
+      expected_active_bits = 8;
+    }
+    SECTION("10-bit Resolution")
+    {
+      expected_active_bits = 10;
+    }
+    SECTION("12-bit Resolution")
+    {
+      expected_active_bits = 12;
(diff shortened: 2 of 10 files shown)


library/L1_Peripheral/lpc40xx/adc.hpp
```diff
@@ -190,6 +190,9 @@ class Adc final : public sjsu::Adc
   /// largest value it can have is 2^12 = 4096
   static constexpr uint8_t kActiveBits = 12;
 
+  /// The LPC40xx ADC has a reference voltage of 3.3V.
+  static constexpr units::voltage::microvolt_t kReferenceVoltage = 3.3_V;
```
kammce, line 194: It would make sense for the reference voltage for the ADC to come from a constructor. Having the parameter have a default value of 3v3 is good, but users of the lpc40xx ADC may choose a different reference voltage than 3v3 depending on their application.

library/L1_Peripheral/test/adc_test.cpp
```diff
@@ -0,0 +1,97 @@
...
+    uint32_t expected_resolution = (1 << expected_active_bits) - 1;
+
+    // Verify
+    Verify(Method(mock_adc, GetActiveBits)).Once();
+    CHECK(actual_resolution == expected_resolution);
+  }
+
+  SECTION("Voltage")
+  {
+    mock_adc.ClearInvocationHistory();
```
kammce, line 48:
```suggestion
```
This is not required as each Test Case and section is its own instance. Meaning, for every section, the code above is reconstructed.

library/L2_HAL/sensors/environment/light/temt6000x01.hpp
```diff
@@ -13,36 +13,29 @@ class Temt6000x01 final : public LightSensor
 {
  public:
   /// @param adc The ADC peripheral used to read the sensor data.
-  /// @param adc_reference_voltage The reference voltage of the ADC. This value
-  ///                              will vary based on the chosen MCU.
   /// @param pull_down_resistance The resistance of the pull down resistor used
   ///                             in the circuit.
   explicit constexpr Temt6000x01(sjsu::Adc & adc,
-                                 units::voltage::volt_t adc_reference_voltage,
```
kammce, line 21: It would be a good idea to keep this in the constructor, but in this case, the reference voltage wouldn't be for the ADC it would be for the supply voltage for the light sensor.

library/L2_HAL/sensors/environment/light/temt6000x01.hpp
```diff
@@ -76,12 +73,9 @@ class Temt6000x01 final : public LightSensor
  private:
   /// The ADC peripheral used to capture the input illuminance data.
   const sjsu::Adc & adc_;
-  /// Reference voltage of the ADC peripheral.
-  const units::voltage::microvolt_t kAdcReferenceVoltage;
+  /// The voltage supplied to this device.
+  const units::voltage::volt_t kReferenceVoltage;
```
kammce, line 77: It doesn't look like you are using this member at all in the class.

library/L2_HAL/sensors/environment/light/temt6000x01.hpp
```diff
@@ -13,36 +13,32 @@ class Temt6000x01 final : public LightSensor
...
  public:
   /// @param adc The ADC peripheral used to read the sensor data.
-  /// @param adc_reference_voltage The reference voltage of the ADC. This value
-  ///                              will vary based on the chosen MCU.
+  /// @param reference_voltage The voltage supplied to this device.
   /// @param pull_down_resistance The resistance of the pull down resistor used
   ///                             in the circuit.
   explicit constexpr Temt6000x01(sjsu::Adc & adc,
-                                 units::voltage::volt_t adc_reference_voltage,
+                                 units::voltage::volt_t reference_voltage,
```
kammce, line 20: A better name may be `supply_voltage`

HuangDave: Would the supply voltage be necessary since it's never used?

kammce: Well yes because bit should be used. Supply voltage should be used in this driver

library/L2_HAL/sensors/environment/light/temt6000x01.hpp
```diff
@@ -13,33 +13,32 @@ class Temt6000x01 final : public LightSensor
...
-  /// @return The illuminance in units of lux ranging from 1 - 1'000.
+
+  /// @returns The illuminance in units of lux ranging from 1 - 1'000.
   units::illuminance::lux_t GetIlluminance() const override
   {
-    // voltage = adc_output * (adc_reference_voltage / adc_resolution)
+    // voltage = reference_voltage * (adc_output / adc_resolution)
     const units::voltage::microvolt_t kVoltage(
-        adc_.Read() * kAdcReferenceVoltage / adc_resolution_);
+        kSupplyVoltage * adc_.Read() / adc_.GetMaximumValue());
```
kammce, line 41: After some thought about this, I realized that I was wrong. If all you need to do for this driver is derive the current, this can be done without the need of a supply voltage variable. We only need the absolute voltage across the pull down resistor to figure out the amount of current passing through the light sensor. So you can revert this back to simply using `adc.Voltage()`.

kammce: Revert the change back and you can submit this.

library/L2_HAL/sensors/environment/light/temt6000x01.hpp
kammce: Realized that the supply voltage is not necessary for this driver whats so ever and the current can be derived without knowing the circuits supply voltage.

PR #1215, opened 2020-05-17
Add Clock Control Documentation

### LINKED ISSUE #1164: HOW TO: Use System Controller Clock Configuration

kammce:
- Changed name of sjtwo's demo to reflect the common name for
  clock_configuration projects.
- Reworded the documentation for the SystemController interface

Resolves #1164


documentation/guides/clock_systems.md
```diff
@@ -0,0 +1,275 @@
...
+if (!result)
+{
+  // Handle error ...
+}
+```
+
+This option seems like the obvious choice but has a fatal issue. During platform
+initialization, peripherals and other drivers that are enabled at startup will
+have adjusted their own clock frequencies based on the frequency they were fed
+initially. If you change these without also rerunning the those driver's
```
HuangDave, line 184: initially. If you change these without also rerunning ~~the~~ those driver's

documentation/guides/clock_systems.md
```diff
@@ -0,0 +1,275 @@
...
+initialize methods, their clocks will be incorrect and they will not performed
```
HuangDave, line 185: initialize methods, their clocks will be incorrect and ~~they~~ **the drivers** will not **perform~~ed~~**

documentation/guides/clock_systems.md
```diff
@@ -0,0 +1,275 @@
...
+sjsu::InitializePlatform();
+```
+
+Calling the `InitializePlatform()` function directly, will prevent the issues in
+Method 1, since `InitializePlatform()` will handle re-initializing the
+peripherals based on the new clock settings.
+
+!!! Note
+    This only re-initializes the peripherals used at startup. These WILL NOT
+    re-initialize all of the peripherals and driver you have in your
```
HuangDave, line 199: re-initialize all of the peripherals and **drivers** you have in your

documentation/guides/clock_systems.md
```diff
@@ -0,0 +1,275 @@
...
+In this case, you either need to re-run the initialization of each of the
+drivers and peripherals in your application or simply set the clocks before
+initialization, which tends to reduce code complexity and size. Failure to do
+this will lead to undefined behavior.
+
+### Changing clock rates in a thread
+First off, DON'T! But if you feel the need to, prefer to have just 1 thread
+manage this, or have a mutex lock usage of this resource. The clock tree is a
+single hardware resource and the configuration object is a single statically
+allocated objects, thus are not thread safe to manipulate without some sort of
```
HuangDave, line 228: allocated **object**, thus are not thread safe to manipulate without some sort of

documentation/guides/clock_systems.md
```diff
@@ -0,0 +1,275 @@
...
+protection.
+
+### Power Savings
+Set the clock rates to the lowest values possible that still fit your
+performance needs. This will very likely require testing of the platform to
+verify that the platform is performing as expected.
+
+After that point, there is no need to change the clock rates further. If you you
```
HuangDave, line 236: After that point, there is no need to change the clock rates further. If you ~~you~~

PR #1224, opened 2020-05-19
Adding SystemController interface design doc

HuangDave:


documentation/design_docs/L1/system_controller.md
```diff
@@ -0,0 +1,154 @@
...
+
+# Overview
+Every SystemController should take a custom `ClockConfiguration` data
+structure reference as an input in their constructor, unless a system does
+not incorporate any form of clock tree.
+
+> See [Clock Control Guide](https://sjsu-dev2.readthedocs.io/en/latest/guides/clock_systems)
+> for more information regarding the use of `ClockConfiguration`.
+
+The SystemController uses the `PeripheralID` helper class to define various
```
HuangDave, line 54: Would it be better to name **PeripheralID** as **ResourceID**?

documentation/design_docs/L1/msp432p401r/system_controller.md
```diff
@@ -12,14 +12,14 @@
     - [void UnlockClockSystemRegister() const](#void-unlockclocksystemregister-const)
     - [void LockClockSystemRegister() const](#void-lockclocksystemregister-const)
   - [Initialization](#initialization)
-    - [Returns\<void> Initialize() override](#returnsvoid-initialize-override)
-    - [Returns\<units::frequency::hertz_t> ConfigureDcoClock() const](#returnsunitsfrequencyhertzt-configuredcoclock-const)
-    - [Returns\<units::frequency::hertz_t> ConfigureReferenceClock() const](#returnsunitsfrequencyhertzt-configurereferenceclock-const)
-    - [Returns\<void> SetClockSource(Clock clock, Oscillator oscillator) const](#returnsvoid-setclocksourceclock-clock-oscillator-oscillator-const)
-    - [Returns\<void> SetClockDivider(Clock clock, ClockDivider divider) const](#returnsvoid-setclockdividerclock-clock-clockdivider-divider-const)
+    - [Returns&lt;void> Initialize() override](#returnsvoid-initialize-override)
```
HuangDave, line 15: Using `&lt;` so that '&lt;' shows up in HTML.

documentation/design_docs/L1/system_controller.md
kammce: Yeah, that sounds like a good generic name for that. Good idea!

documentation/design_docs/L1/msp432p401r/system_controller.md
kammce: Awesome! I was wondering what we could do to make this show up.

documentation/design_docs/L1/system_controller.md
```diff
@@ -0,0 +1,151 @@
...
+
+  template <class ClockConfiguration>
+  ClockConfiguration & GetClockConfiguration();
+};
+```
+
+## Configuring the Default System Controller for the Platform
+
+### static void SetPlatformController(SystemController * system_controller)
+Sets the specified system controller for the platform.
```
kammce, line 106:
```suggestion
Sets the specified system controller for the platform. Each platform is responsible for calling this function correctly from within the `InitializePlatform()` function. Without doing so, access to `GetPlatformController()` will return a nullptr.
```

documentation/design_docs/L1/system_controller.md
```diff
@@ -0,0 +1,151 @@
...
+
+### static SystemController & GetPlatformController()
+Returns the current system controller for the platform.
```
kammce, line 109:
```suggestion
Returns the current system controller for the platform. Note that this is used by peripheral drivers to power themselves on and determine their running clock rate. This is also used for application developers to set up "...". This should not be used by L2 HAL or L3 Application as knowledge about the particular platform is not and should not be knowable at those levels.
```
Or something along those lines.

documentation/design_docs/L1/system_controller.md
```diff
@@ -0,0 +1,151 @@
...
+
+## Clock Configuration
+
+### void * GetClockConfiguration()
+Returns a generic pointer to the system controller's `ClockConfiguration`.
```
kammce, line 114:
```suggestion
Returns a pointer to the memory location of system controller's `ClockConfiguration`. In generally this should be called directly. Prefer to call `ClockConfiguration & GetClockConfiguration()` instead
```

PR #1295, opened 2020-07-03
Fixing typo in make debug

HuangDave:


kammce:
PR #1294 actually has a fix for this. There is one more typo in make debug than what is shown here.

Issue #1034, opened 2020-01-15
EEPROM driver and demo does not work


kammce:
I get the following output from telemetry

![image](https://user-images.githubusercontent.com/4549768/72440939-9f763880-375e-11ea-8c88-b5b055d271cc.png)

Further down you see this though:

![image](https://user-images.githubusercontent.com/4549768/72440983-b4eb6280-375e-11ea-8234-ebaf8c13468c.png)

So it seems that the addressing is not correct.

harmandersihra:
Just submitted a PR for this issue.

kammce:
Resolved by #1304

PR #1304, opened 2020-07-09
Fixed EEPROM driver

harmandersihra:
Turns out there was an exclamation mark in the line checking the status bit for the program operation, and we were using the wrong bit to clear the status register too. Also it turns out we could drop the max timeout down to 5ms after that change.

There was an issue with program() not being called at the right time, so line 175 had to be changed. It calls it one iteration of the loop earlier.

There were also a couple of typos.


harmandersihra:
This should fix #1034 as well. Trying to link it to this PR.

PR #1318, opened 2020-07-23
Adding initial BLE interface

HuangDave:


library/L2_HAL/communication/bluetooth/bluetooth_device.hpp
```diff
@@ -0,0 +1,45 @@
...
+namespace bluetooth
+{
+/// The interface should be implemented by bluetooth devices that typically
+/// utilize the CC245x series of SoC for Bluetooth Low Energy (BLE)
+/// communication.
+///
+/// The CC245x MCUs provide BLE v4.0 which allows for communication with modern
+/// devices such as iOS or Android devices. These devices typically use UART
+/// communicate with the MCU.
+class BluetoothDevice
```
kammce, line 18: This Interface doesn't really define all bluetooth devices. Or even most. Just bluetooth devices that act as wireless UART transmitters or (transparent UART). It may be a good idea to really think about the bluetooth module as such, an sjsu::Uart module. Now if you want to have additional interface methods specifically for UartBluetooth devices, then having this sort of interface is good. But you don't want to duplicate the work. So imagine this:

```C++
class HM10 : public sjsu::Uart, sjsu::UartOverBluetoothDevice
{
 public:
  virtual void Disconnect() const { /* ... */ } // from the sjsu::UartOverBluetoothDevice
};
```

sjsu::UartOverBluetoothDevice is just an example name. You can come up with whatever you think makes sense. But the main point is that it would be helpful to put the HM10 (and other devices) in places of where a UART port would go.

See esp8266.hpp for another example of a class that implements two interfaces

library/L2_HAL/communication/bluetooth/serial_bluetooth.hpp
```diff
@@ -0,0 +1,35 @@
...
+                           const char * parameter           = nullptr,
+                           std::chrono::nanoseconds timeout = 10ms) const = 0;
+
+  /// Sends a command to the device.
+  ///
+  /// @param command The command to send.
+  /// @param parameters Parameters to go with command if applicable.
+  virtual void SendCommand(const char * command,
+                           const std::initializer_list<uint8_t> parameters = {},
+                           std::chrono::nanoseconds timeout = 10ms) const = 0;
```
kammce, line 32:
```suggestion
  void SendCommand(const char * command,
                           const std::initializer_list<uint8_t> parameters = {},
                           std::chrono::nanoseconds timeout = 10ms) const = 0;
```

This doesn't need to be `virtual`. You can have this version of SendCommand() call the virtual version.

library/L2_HAL/communication/bluetooth/serial_bluetooth.hpp
```diff
@@ -0,0 +1,35 @@
...
+/// serial communication such as SPI or UART for the transferring of data.
+class SerialBluetooth
+{
+ public:
+  /// @returns True if the device currently has a connection with a another
+  ///          device.
+  virtual bool GetState() const = 0;
+
+  /// Disconnect from the current connection.
+  virtual void Disconnect() const = 0;
```
kammce, line 20: What about `Connect()`?

library/L2_HAL/communication/bluetooth/serial_bluetooth.hpp
```diff
@@ -0,0 +1,35 @@
...
+
+  virtual void SendCommand(const char * command,
+                           const char * parameter           = nullptr,
+                           std::chrono::nanoseconds timeout = 10ms) const = 0;
```
kammce, line 24: We may want to brain storm on this. I'm not sure how useful this interface is at the moment compared to just a straight driver. We should meetup next week to discuss this.

kammce:
Closing this for now. We can open this back up when we think its a good time to figure out BLE and how to make a good interface for it in SJSU-Dev2.

PR #1320, opened 2020-07-23
Fix .data and .bss sections issue with unit tests

kammce:
Migrate dummy Ram definitions to main_test.cpp in order to make sure
that it is defined for all builds.


kammce:
@HuangDave you can merge this if it completes CI.

PR #1363, opened 2020-09-08
[WIP] Make all interfaces adopt Returns<>

kammce:
It can not be determined that that any particular implementation of a
hardware driver will or will not throw an error, thus every interface
must be equip to return an error if need be.


kammce:
It was decided to transition to exception handling rather than return error types. For large enough projects exceptions are smaller in code size and faster when exceptions are not running.

PR #1364, opened 2020-09-12
Add max17043 driver and SOC interface.

AndrewDEmerson:
Add an interface for devices that return the state of charge of a device.
Also add the max17043 battery charge driver that uses the interface.


kammce:
Squash your commit into 1 commit. This should fix the issues with tidy.

PR #1384, opened 2020-10-22
Fix homebrew check in setup

kammce:
This should prevent homebrew from being reinstalled when it already
exists.


kammce:
Resolved by #1395

PR #1410, opened 2020-12-07
Fix Code Coverage Percentage

kammce:


kammce:
This change eliminates the coverage of unit test files themselves as they are not indicative of the actual coverage of the main libraries.

kammce:
Wasn't able to eliminate the partials issue. Will attempt again in a later PR.

PR #1411, opened 2020-12-08
PWM Driver

harmandersihra:
So, I'll do this here:

The STM has a bunch of different timers, and it also has multiple channels for most of the timers. Some timers however, only support channel 1. So to simplify, I only used channel 1 and made each PWM option a different timer. I looked at the alternate function settings to see which pins corresponded to which timers on which channels, and I chose pins that were on channel 1 for the timers with NO remapping required. If you look at the timer structs in pwm.h, you'll see some have Y and some have N next to them. The ones with Y were the ones I saw working waveforms for on the scope.

So it seems like it's not an issue with it being on different ports because I have pins on port A and B working. All of the working timers are enabled through RCC_APB1ENR, so I was thinking that might be an issue too.

If you wanted to help debug, I think we should get Timer 1 working since well, it's the first timer, and it's on RCC_APB2ENR.


library/L1_Peripheral/stm32f10x/pwm.hpp
```diff
@@ -0,0 +1,296 @@
...
+  struct PwmPin_t
+  {
+    // Reference to the pin
+    sjsu::Pin & pin;
+
+    // Holds the address to the corresponding Timer peripheral
+    TIM_TypeDef * registers;
+
+    // Holds the channel (1 - 4) which corresponds to the pin
+    uint8_t channel : 3;
```
kammce, line 30:
```suggestion
    uint8_t channel;
```
Remove the bit field. It can cause really annoying warning and errors and it doesn't save us any space without making the struct packed. Also don't make the struct packed because you end up wasting space in the generated code to pull out those 3 bits than you would get back form packing this struct.

library/L1_Peripheral/stm32f10x/pwm.hpp
```diff
@@ -0,0 +1,296 @@
...
+    inline static sjsu::stm32f10x::Pin pb1 = sjsu::stm32f10x::Pin('B', 1);
+
+    // Timer 4
+    inline static sjsu::stm32f10x::Pin pb6 = sjsu::stm32f10x::Pin('B', 6);
+    inline static sjsu::stm32f10x::Pin pb7 = sjsu::stm32f10x::Pin('B', 7);
+    inline static sjsu::stm32f10x::Pin pb8 = sjsu::stm32f10x::Pin('B', 8);
+    inline static sjsu::stm32f10x::Pin pb9 = sjsu::stm32f10x::Pin('B', 9);
+
+   public:
+    inline static const PwmPin_t kPin_a_8 = {
```
kammce, line 65:
```suggestion
    inline static const PwmPin_t kA8 = {
```
This is a more concise name for the pin that still makes it obvious what you are doing.

library/L1_Peripheral/stm32f10x/pwm.hpp
```diff
@@ -0,0 +1,296 @@
...
+    TIM_TypeDef * registers;
+
+    // Holds the channel (1 - 4) which corresponds to the pin
+    uint8_t channel : 3;
+
+    // Peripheral ID for the timer
+    sjsu::SystemController::ResourceID id;
+  };
+
+  struct PwmPin
```
kammce, line 36:
```suggestion
  struct Channel
```

library/L1_Peripheral/stm32f10x/pwm.hpp
kammce: Although I think I need a new API for PWM in general to make it more explicit which pins go to which timers so that users don't have to go to the datasheet to know which pins go to which timers.

library/L1_Peripheral/stm32f10x/pwm.hpp
```diff
@@ -0,0 +1,296 @@
...
+    }
+
+    // Enables the output channel
+    uint8_t kOffset = static_cast<uint8_t>(4 * (pwmPin_.channel - 1));
+    pwmPin_.registers->CCER |= static_cast<uint16_t>(1 << kOffset);
+
+    // Enable timer
+    bit::Register(&pwmPin_.registers->CR1)
+        .Set(kTimerEnable)
+        .Save();
```
kammce, line 267: Move this to ModuleEnable(). Also add a means to disable a channel.

library/L1_Peripheral/stm32f10x/pwm.hpp
```diff
@@ -0,0 +1,296 @@
...
+    SystemController::GetPlatformController().PowerUpPeripheral(pwmPin_.id);
+
+    // Set pin to alternative function to support PWM
+    pwmPin_.pin.ConfigureFunction(1);
+
+    // Here we set the PWM mode and the preload enable
+    if (pwmPin_.channel < 3)
+    {
+      pwmPin_.registers->CCMR1 |= 
+        static_cast<uint16_t>(0b1101 << (((pwmPin_.channel - 1) * 8) + 3));
```
kammce, line 246: Can you make constants for these magic numbers. Why 8? Why +3?

kammce (changes requested):

harmandersihra:
This should be the final implementation. All presubmit checks pass and all pins are functional.

library/L1_Peripheral/stm32f10x/pwm.hpp
```diff
@@ -0,0 +1,446 @@
...
+    TIM_TypeDef * registers;
+
+    // Holds the channel (1 - 4) which corresponds to the pin
+    uint8_t channel;
+
+    // Peripheral ID for the timer
+    sjsu::SystemController::ResourceID id;
+  };
+
+  struct Channel_t
```
kammce, line 32:
```suggestion
  struct Channel  // NOLINT
```

Its an unwritten rule in SJSU-Dev2 to remove the `_t` for namespaces like this inside of a class. Its how the other classes are formatted so this should follow.

library/L1_Peripheral/stm32f10x/pwm.hpp
```diff
@@ -0,0 +1,446 @@
...
+      pwm_pin_.registers->CCR4 =
+        static_cast<uint16_t>(kClampedDutyCycle * pwm_pin_.registers->ARR);
+      break;
+    }
+  }
+
+  float GetDutyCycle() override
+  {
+    return
+      static_cast<float>(pwm_pin_.registers->CCR1 / pwm_pin_.registers->ARR);
```
kammce, line 229:
```suggestion
    auto ccr1 = static_cast<float>(pwm_pin_.registers->CCR1);
    auto arr = static_cast<float>(pwm_pin_.registers->ARR);
    return crr1 / arr;
```

You are probably wondering why this is necessary. Because if you divide two ints you get a none float number. meaning, if your value is below 1 it will truncate to 0. So we need to cast both of these into float so we can perform floating point division not int division. Also I gave the two variables the name `ccr1` and `arr`, can you give them variable names that represent what they really are.

library/L1_Peripheral/stm32f10x/pwm.hpp
```diff
@@ -0,0 +1,446 @@
...
+
+    if (enable)
+    {
+      // Preload enable
+      switch (pwm_pin_.channel)
+      {
+      case 1:
+        bit::Register(&pwm_pin_.registers->CCMR1)
+          .Set(kPreloadEnableLow)
+          .Save();
```
kammce, line 258:
```suggestion
        bit::Register(&pwm_pin_.registers->CCMR1)
          .Set(kPreloadEnableLow)
          .Set(kChannel1OutputEnable)
          . // etc
          .Save();
```

You can chain these together. The bit::Register() class is efficient if you chain a bunch of modifications together and then at the end `Save()` the final result. That way the computation can occur mostly in registers, sometimes in cache, worst case using RAM. But the benefit is that it doesn't have to access all the way into peripherals for every statement. Also chaining these calls is actually constexpr, so a lot of this turns into a single load instruction, but the code still looks very expressive.

This also eliminates the multiple uses of switch case where you may be able to get it down to just 1.

library/L1_Peripheral/stm32f10x/pwm.hpp
```diff
@@ -0,0 +1,446 @@
...
+        bit::Register(&pwm_pin_.registers->BDTR)
+          .Clear(kMainOutputEnable)
+          .Save();
+      }
+
+      // Timer enable
+      bit::Register(&pwm_pin_.registers->CR1)
+        .Clear(kTimerEnable)
+        .Save();
+    }
```
kammce, line 379: Wait does this make sense though? If I want to disable just 1 channel, why do I have to turn off the whole timer and also lose all of my other channels. This doesn't seem like the behavior you'd expect from disabling a single channel of PWM.

library/L1_Peripheral/stm32f10x/pwm.hpp
```diff
@@ -0,0 +1,446 @@
...
+  }
+
+  void ModuleInitialize() override
```
kammce, line 382: Put `ModuleInitialize()` above `ModuleEnable()` in terms of code order.

library/L1_Peripheral/stm32f10x/test/pwm_test.cpp
```diff
@@ -0,0 +1,161 @@
...
+    test_subject.ConfigureFrequency(frequency);
+
+    // Verify
+    CHECK(prescalar == local_timer.PSC);
+    CHECK(divider == local_timer.ARR);
+  }
+
+  SECTION("SetDutyCycle()")
+  {
+    float duty_cycle;
```
kammce, line 112:
```suggestion
    // Setup
    float duty_cycle;
```

NIT: Add `// Setup` here.

kammce (changes requested):
Lookin' great Harmander! A few things to clean up and this will be ready to be uploaded. Thank you for your hard work!

library/L1_Peripheral/stm32f10x/pwm.hpp
harmandersihra: Just noticed this was wrong since we're doing multi channel now. I'll rewrite this a bit

library/L1_Peripheral/stm32f10x/pwm.hpp
harmandersihra: Unfortunately this won't work because all the registers are different. One switch affects CCMRx, the other does CCER, and then BDTR and CR1. Unless I'm misunderstanding how the chaining works.

library/L1_Peripheral/stm32f10x/pwm.hpp
harmandersihra: Yea, that's a good point. I'll fix that.

library/L1_Peripheral/stm32f10x/pwm.hpp
harmandersihra: So one option to get rid of the switches would be to store witch CCRx register each PwmPin_t object uses. CCMRx is little more tricky since we'd have to store the bit mask as well but that way we could get rid of the channel entry in the struct too. So overall we'd add one entry to the struct and get rid of all switches. Thoughts?

harmandersihra: We would actually still need the switch for CCER at that point unless we add another bit mask entry.

kammce: Ah ok I missed that. Still doesn't explain why there are two switch case blocks if both of them use the same channel variable. You should group both statements under the same case.

kammce (changes requested):

kammce (approved):

PR #1448, opened 2021-03-29
Added compare_floats function and utility test

SarahS16:


library/utility/math/comparef.hpp
```diff
@@ -0,0 +1,28 @@
...
+        return false;
+    }
+
+
+}
+
+
+
+
+
```
kammce, line 27: Please reduce the number of newlines

kammce (changes requested):

PR #1452, opened 2021-11-13
Initial version of Canbus for the stm32f10x

MasterBoyd:


library/peripherals/stm32f10x/can.hpp
```diff
@@ -0,0 +1,763 @@
...
+    return can1;
+  }
+  else
+  {
+    static_assert(InvalidOption<port>,
+                  SJ2_ERROR_MESSAGE_DECORATOR(
+                      "Support CAN ports for STM32F10xx are CAN1"));
+    return GetCan<0>();
+  }
+}
```
kammce, line 730:
```suggestion
inline Can & GetCan()
{
    static auto & port_transmit_pin = GetPin<'A', 12>();
    static auto & port_read_pin     = GetPin<'A', 11>();

    /// Predefined definition for CAN
    static const Can::Port_t kCan = {
      .td_pin = port1_transmit_pin,
      .rd_pin = port1_read_pin,
      .can    = stm32f10x::CAN1,
      .id     = sjsu::stm32f10x::SystemController::Peripherals::kCan1,
    };

    static Can can(kCan);
    return can;
}
```

If there is only 1 can then you don't need to take a template parameter.

library/peripherals/stm32f10x/can.hpp
```diff
@@ -0,0 +1,763 @@
...
+typedef struct
+{
+  uint16_t id;
+  uint8_t data[8];
+  uint8_t len;
+} CAN_msg_t;
+
+void CANReceive(CAN_msg_t * CAN_rx_msg)
```
kammce, line 738:
```suggestion
inline void CANReceive(CAN_msg_t * CAN_rx_msg)

```

if this is going to in the header file and outside of a class it needs `inline`. Also why is this here?

library/peripherals/stm32f10x/can.hpp
```diff
@@ -0,0 +1,763 @@
...
+                      "Support CAN ports for STM32F10xx are CAN1"));
+    return GetCan<0>();
+  }
+}
+typedef struct
+{
+  uint16_t id;
+  uint8_t data[8];
+  uint8_t len;
+} CAN_msg_t;
```
kammce, line 736: Why do you have this?

library/peripherals/stm32f10x/can.hpp
```diff
@@ -0,0 +1,763 @@
...
+  CAN_rx_msg->data[6] = 0xFFUL & (stm32f10x::CAN1->sFIFOMailBox[0].RDHR >> 16);
+  CAN_rx_msg->data[7] = 0xFFUL & (stm32f10x::CAN1->sFIFOMailBox[0].RDHR >> 24);
+
+  stm32f10x::CAN1->RF0R |= 0x20UL;
+}
+
+uint8_t CANMsgAvail(void)
+{
+  return CAN1->RF0R & 0x3UL;
+}
```
kammce, line 760: What is this for? Also don't use C style `function(void)` remove the void.

kammce (changes requested):

